PR #1, opened 2017-06-08
Add connect_timeout

sfackler:
Unlike the rest of the API, connect_timeout does not correspond to a
single API call. However, it's a fundamental operation, and is a pain to
set up.

### DIFF
--- a/src/sys/unix/mod.rs
+++ b/src/sys/unix/mod.rs
@@ -18,7 +18,7 @@ use std::net::{self, Ipv4Addr, Ipv6Addr, SocketAddrV4, SocketAddrV6, SocketAddr}
 use std::ops::Neg;
 use std::os::unix::prelude::*;
 use std::sync::atomic::{AtomicBool, Ordering, ATOMIC_BOOL_INIT};
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use libc::{self, c_void, c_int, sockaddr_in, sockaddr_storage, sockaddr_in6};
 use libc::{sockaddr, socklen_t, AF_INET, AF_INET6, ssize_t};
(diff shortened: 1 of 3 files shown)


sfackler:
Untested on Windows as I don't have access to a Windows box for a couple days.

src/sys/unix/mod.rs
```diff
@@ -118,6 +118,53 @@ impl Socket {
...
+        let mut timeout = timeout.as_secs()
+            .saturating_mul(1_000)
+            .saturating_add(timeout.subsec_nanos() as u64 / 1_000_000);
+        if timeout == 0 {
+            timeout = 1;
+        }
+
+        let timeout = cmp::min(timeout, c_int::max_value() as u64) as c_int;
+
+        let r = unsafe { libc::poll(&mut pollfd, 1, timeout) };
```
alexcrichton: I think this'll want to loop to handle `EINTR`, right? (or at least that'd what I'd expect a "convenient" implementation to do

src/sys/unix/mod.rs
```diff
@@ -118,6 +118,53 @@ impl Socket {
...
+            -1 => return Err(io::Error::last_os_error()),
+            0 => return Err(io::Error::new(io::ErrorKind::TimedOut, "connection timed out")),
+            _ => {
+                if let Some(e) = self.take_error()? {
+                    return Err(e);
+                }
+            }
+        }
+
+        self.set_nonblocking(false)
```
alexcrichton: Perhaps this could be lifted above the `match r` to unconditionally happen?

src/sys/unix/mod.rs
sfackler: What should we do about the timeout in that situation? Just use the original one, or try to keep it updated with a couple of Instants?

Doesn't everything use SA_RESTART, anyway?

sfackler:
Updated

alexcrichton:
👍

PR #2, opened 2017-06-12
Add a SockAddr type

sfackler:
We don't want to be stuck only supporting IP sockets

### DIFF
--- a/src/sockaddr.rs
+++ b/src/sockaddr.rs
@@ -4,9 +4,11 @@ use std::ptr;
 use std::fmt;
 
 #[cfg(unix)]
-use libc::{sockaddr, sockaddr_storage, sa_family_t, socklen_t, AF_INET, AF_INET6};
+use libc::{sockaddr, sockaddr_storage, sockaddr_in, sockaddr_in6, sa_family_t, socklen_t, AF_INET,
+           AF_INET6};
 #[cfg(windows)]
 use winapi::{SOCKADDR as sockaddr, SOCKADDR_STORAGE as sockaddr_storage,
+             SOCKADDR_IN as sockaddr_in, SOCKADDR_IN6 as sockaddr_in6,
              ADDRESS_FAMILY as sa_family_t, socklen_t, AF_INET, AF_INET6};
 
 use SockAddr;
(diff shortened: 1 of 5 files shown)


src/sockaddr.rs
```diff
@@ -0,0 +1,130 @@
...
+            len: len,
+        }
+    }
+
+    unsafe fn as_<T>(&self, family: sa_family_t) -> Option<T> {
+        if self.storage.ss_family != family {
+            return None;
+        }
+
+        Some(mem::transmute_copy(&self.storage))
```
alexcrichton: One day this may bite us, but let's hope it's way far away!

src/sockaddr.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /// Returns this address as a `SocketAddrV4` if it is in the `AF_INET`
+    /// family.
+    pub fn as_inet(&self) -> Option<SocketAddrV4> {
```
alexcrichton: Perhaps this could return `&SocketAddrV4` instead? (a relatively large struct to copy as we go down)

src/sockaddr.rs
```diff
@@ -0,0 +1,130 @@
...
+        &self.storage as *const _ as *const _
+    }
+}
+
+// SocketAddrV4 and SocketAddrV6 are just wrappers around sockaddr_in and sockaddr_in6
+
+impl From<SocketAddrV4> for SockAddr {
+    fn from(addr: SocketAddrV4) -> SockAddr {
+        unsafe {
+            SockAddr::from_raw_parts(&addr as *const _ as *const _,
```
alexcrichton: Could this throw in some `assert_eq!` about the size of `SocketAddrV4`? (same for v6 below)

src/sockaddr.rs
sfackler: I was somewhat terrified when I saw that this was how the `addr2raw` function was implemented as well, but it works.

src/sockaddr.rs
sfackler: That will lock us into guaranteeing that `SocketAddrV4` has the same representation as `sockaddr_in`, but if that's okay then sure.

alexcrichton: Oh wait good point, let's not do that.

sfackler:
`to_socket_addr` may not be the best name since there are both `std::net::SocketAddr` and `std::os::unix::net::SocketAddr` :(

src/sockaddr.rs
sfackler: Done via `transmute` to check statically.

alexcrichton:
Ok, let's stick with it as-is and fix up ergonomics as we see fit. Thanks again!

(if you'd like a publish of this just lemme know)

alexcrichton:
Er actually, added you as collaborator + crate owner, feel free to publish if you need it! (just don't forget to bump the version)

sfackler:
Thanks!

PR #4, opened 2017-07-08
Unix socket support

sfackler:


alexcrichton:
Looks great! Mind adding a test or two exercising this?

Also, can you change `Cargo.toml` to render these features by default on docs.rs, and also mention in the documentation the restrictions on when they're available?

sfackler:
Updated

alexcrichton:
👍

PR #9, opened 2017-09-24
Add protocol 'constants'

DoumanAsh:
I'd prefer to have constants as actual constants instead of creating it through functions but I decided to follow style as with Domain.

Due to difference in how these constants defined in winapi and libc I put implementation in sys module as otherwise I would just need to re-define all these constants in sys module...

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -762,6 +762,28 @@ impl Type {
     }
 }
 
+impl ::Protocol {
+    /// Protocol corresponding to `ICMPv4`
+    pub fn icmpv4() -> Self {
+        ::Protocol(sys::IPPROTO_ICMP)
+    }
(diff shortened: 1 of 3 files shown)


alexcrichton:
Could these be added at the top level instead of in the platform specific modules? Otherwise looks great, thanks!

DoumanAsh:
@alexcrichton They can, but as i mentioned i would need to have unified re-definition of constants, if you don't mind.

alexcrichton:
Yes this'll require a different organization, but that's ok

DoumanAsh:
Updated. Took a bit of change in windows code as it used old struct constants for protocols

alexcrichton:
Thanks!

PR #10, opened 2017-10-14
Fix refused connections in connect_timeout

sfackler:


alexcrichton:
OSX/Windows seem unhappy?

sfackler:
Should be fixed.

PR #12, opened 2017-12-23
Update winapi dep to 0.3, and update code to support it

keeperofdakeys:
This PR adds support for winapi 0.3 to socket2-rs. A few things to keep in mind:

- Since this library exposes winapi types in public interfaces, the version should probably be bumped to 0.3 as upgrading may be a breaking change.
- Currently this doesn't build with the winapi 0.3.2, as the `SOCKADDR_IN6_LH` struct is only in git.. This does build (and passes tests) on windows when compiled with the below additions to the winapi dep in Cargo.toml:
```
git = "https://github.com/retep998/winapi-rs.git"
branch = "0.3"
```
- The `SOCKET` type in winapi is now `usize` instead of `u64`, which doesn't match the `std::os::windows::io::RawSocket` type in Rust that's still `u64`. So this PR casts between them in a few places.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,10 +15,9 @@ possible intended.
 [package.metadata.docs.rs]
 all-features = true
 
-[target."cfg(windows)".dependencies]
-ws2_32-sys = "0.2"
-winapi = "0.2"
-kernel32-sys = "0.2"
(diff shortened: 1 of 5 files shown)


alexcrichton:
Thanks! I'm down to merge as soon as CI is green.

PR #19, opened 2018-03-13
fix ipv6 proto reference for join ipv6 multicast join

bluejekyll:
After upgrading to `socket2-rs` (which does clean up some code paths compared to `net2`, thanks for the recommendation) I noticed that `join_multicast_v6` and `leave_multicast_v6` were not referencing the correct protocol (causes an error on registration). This fixes that.

btw, this also includes automated rustfmt changes... let me know if you want me to revert that.

Also, do you want me to bump the version for release?

### DIFF
--- a/src/sys/unix/mod.rs
+++ b/src/sys/unix/mod.rs
@@ -10,7 +10,7 @@
 
 use std::cmp;
 use std::fmt;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oops thanks! And no worries I'll publish a new version with this in it, thanks!

PR #21, opened 2018-03-20
Fix newtype conversions to i32

0xd34d10cc:
`i32::from(Protocol::tcp())` currently causes stack overflow. Oops!


alexcrichton:
Oh dear, thanks!

PR #25, opened 2018-05-25
Redox OS support

jD91mZM2:
Might be worth releasing a libc update first before to avoid using the git version

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -15,7 +15,7 @@ use std::time::Duration;
 #[cfg(all(unix, feature = "unix"))]
 use std::os::unix::net::{UnixDatagram, UnixListener, UnixStream};
 
-#[cfg(unix)]
+#[cfg(any(unix, target_os = "redox"))]
 use libc as c;
 #[cfg(windows)]
 use winapi::shared::ws2def as c;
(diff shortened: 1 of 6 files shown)


alexcrichton:
Since there's a lot of `#[cfg]` going on here, can redox be moved to a separate file like windows?

jD91mZM2:
That's probably a good idea, although I'm a little worried it'll be a lot of repeated code anyway. The unix file is 1000 lines long and eventually redox will gain support for most of this I assume.

jD91mZM2:
Alright, there we go. It would be cool if I could somehow reuse the functions that are the same on both, but I assume rust wouldn't like `use`ing the unix file on redox.

jD91mZM2:
@alexcrichton Bump?

src/socket.rs
```diff
@@ -827,23 +827,27 @@ impl Type {
         Type(c::SOCK_DGRAM)
     }
 
+    #[cfg(not(target_os = "redox"))]
```
alexcrichton: This is a public API and so I'd prefer to not change the public interface of this crate depending on the platform currently. Could the redox module use a sentinel value for this constant which causes a runtime error when used in this crate?

jD91mZM2: Good point! Updated :)

alexcrichton:
Er sorry so my goal here is to reduce the amount of `#[cfg]` here as much as possible because it otherwise adds maintenance burden to the rest of the code on other platforms. Instead of panicking could a runtime error be issued only when a socket is attempted to be created with an invalid protocol not supported?

alexcrichton:
Thanks!

jD91mZM2:
Awesome! Would be great if you could bump the version of this & libc. Sorry for being annoying.

alexcrichton:
I've published these changes and the libc changes are happening in the libc repo

PR #32, opened 2019-05-05
Fix Socket::nodelay() for Windows

YorickPeterse:
This fixes `Socket::nodelay()` and `Socket::set_nodelay()` for Windows by using a `c_char` instead of a `c_int`, ensuring the type size assertion in `getsockopt` does not fail.

This fixes https://github.com/alexcrichton/socket2-rs/issues/31

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -982,4 +982,16 @@ mod test {
         #[cfg(unix)]
         assert_eq!(socket.keepalive().unwrap(), None);
     }
+
+    #[test]
+    fn nodelay() {
+        let socket = Socket::new(Domain::ipv4(), Type::stream(), None).unwrap();
+
+        assert!(socket.set_nodelay(true).is_ok());
+
+        let result = socket.nodelay();
+
+        assert!(result.is_ok());
+        assert!(result.unwrap());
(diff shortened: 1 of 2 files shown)


YorickPeterse:
@alexcrichton Any feedback on the changes? Thanks in advance for taking a look!

alexcrichton:
Hm so according to [the Windows documentation](https://docs.microsoft.com/en-us/windows/desktop/api/winsock/nf-winsock-setsockopt) the type for `TCP_NODELAY` is BOOL which corresponds to `c_int`?

alexcrichton:
Hm well in any case matching libstd seems reasonable here and I guess this is a bug in the windows documentation!

YorickPeterse:
@alexcrichton It indeed is a bit weird Windows returns a differently sized type here. Perhaps it's something that was introduced after the documentation was written, or the documentation is simply wrong.

Would it be possible to push a new version to crates.io containing the latest `master`? Thanks!

alexcrichton:
Sure thing, published now.

YorickPeterse:
@alexcrichton Thanks! :fireworks:

PR #36, opened 2019-07-26
Fix compilation on illumos if reuseport feature is enabled

andykilroy:
On illumos the socket2-rs project wouldn't compile if the reuseport feature was requested.  This pull request conditionally compiles uses of the SO_REUSEPORT constants so that if an application is written to call set_reuseport(...) or reuseport() socket2-rs will compile on a Solaris based platform.

I had problems testing what was on master (v0.3.10) because there were compilation problems due to cfg_if, so I've based these corrections on 0.3.9, a version I could build and test.  I'm aware files have moved around; I'm mainly sending the pull request in this form to find out how you'd like to proceed.

Note my tests were manual.  I couldn't think of a good strategy to write unit/integration tests for this.


andykilroy:
Issue https://github.com/alexcrichton/socket2-rs/issues/37  details the problem I encountered with 0.3.10

alexcrichton:
Could this perhaps be a bit less intrusive than sprinkling `#[cfg]` wherever it can? I don't think these functions should exist on solaris if solaris doesn't have the functionality.

andykilroy:
Removed my additional #cfg lines.  Placed the not(target_os = "solaris") directives on the #cfg macros that were already there.

alexcrichton:
Looks good to me, but looks like there's a merge conflict as well?

andykilroy:
Merged master into my branch (that was originally based off of 0.3.9).

PR #40, opened 2019-08-07
Fix invalid use of accept4() system call

YorickPeterse:
The accept4 system call is supposed to be given a pointer to an address
length, and not a length as an int. Doing so will lead to an EFAULT
error.

This fixes https://github.com/alexcrichton/socket2-rs/issues/38


alexcrichton:
Oops, thanks!

PR #41, opened 2019-09-06
Add packet domain (AF_PACKET)

drrlvn:


nickrtorres:
This would be really nice to have!
Although, I think this feature should be marked for "linux" instead of "unix", since (AFAIK), `AF_PACKET` is a linux specific feature. On my *nix system (macOS 10.14.5; Darwin Kernel Version 18.6.0) `socket(2)` only shows the following domains:
```
           PF_LOCAL        Host-internal protocols, formerly called PF_UNIX,
           PF_UNIX         Host-internal protocols, deprecated, use PF_LOCAL,
           PF_INET         Internet version 4 protocols,
           PF_ROUTE        Internal Routing protocol,
           PF_KEY          Internal key-management function,
           PF_INET6        Internet version 6 protocols,
           PF_SYSTEM       System domain,
           PF_NDRV         Raw access to network device
```

alexcrichton:
Thanks for the PR! Looks like this doesn't exist on OSX though?

drrlvn:
@nickrtorres, @alexcrichton I set the target OS to Linux, don't know if leaving `unix` in there makes sense or not.

alexcrichton:
Thanks!

PR #43, opened 2019-09-19
Add as_socket_addr convenience method

jean-airoldie:
This adds `as_socket_addr` with tries to convert the `SockAddr` into a `SocketAddr` if possible.

I used the same naming convention as the other `as_inet` methods.


alexcrichton:
Thanks! Could this perhaps be called `as_std` to avoid too much naming confusion? Also could the `None` case be documented as to why and when it returns `None`?

jean-airoldie:
Sure thing.

PR #50, opened 2020-01-13
Deduplicate sys::redox with sys::unix, as redox is a unix OS

coolreader18:


Cargo.toml
```diff
@@ -21,7 +21,7 @@ features = ["handleapi", "ws2def", "ws2ipdef", "ws2tcpip", "minwindef"]
 
 [target."cfg(any(unix, target_os = \"redox\"))".dependencies]
 cfg-if = "0.1.6"
-libc = "0.2.42"
+libc = "0.2"
```
alexcrichton: How come this was updated?

coolreader18: I think there was some libc API that was added for Redox only recently, or it might have been that Redox was only made a Unix in libc [in a version after 0.2.42](https://github.com/rust-lang/libc/commit/5408b1079cda9700bd6efe29c374af862d897890). I could check what it was exactly later today if you'd like.

alexcrichton: Ah ok, makes sense.

PR #51, opened 2020-01-18
Support Out-of-Band related functionality

oskardrums:
This PR adds the ability to get/set SO_OOBINLINE flag for `Socket`s via the `set_out_of_band_inline`/`out_of_band_inline` methods.
Also, this PR exposes the `flags` argument of `send`/`recv` etc. methods to allow caller to specify arbitrary flags, including the `MSG_OOB` when needed.

This feature allows TCP sockets utilizing `socket2` to comply with RFC6093 regarding Urgent data etc., and allows for other uses of `send`\`recv` flags as well.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -20,6 +20,11 @@ use libc as c;
 #[cfg(windows)]
 use winapi::shared::ws2def as c;
 
+#[cfg(any(unix, target_os = "redox"))]
+use libc::MSG_OOB;
+#[cfg(windows)]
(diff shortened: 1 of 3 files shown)


oskardrums:
@alexcrichton review? :)

alexcrichton:
Thanks for the PR! 

Would it be possible to avoid updating the public API of this crate? Could the new methods which support flags be added as separate methods?

alexcrichton:
Thanks!

PR #57, opened 2020-03-29
Move Domain

Thomasdezeeuw:
Moves `Domain` impl to where the type is defined  and moves Unix only `Domain` methods to unix file.


alexcrichton:
👍

PR #59, opened 2020-03-30
Move Type impl where its defined

Thomasdezeeuw:
Also add Type::{non_blocking, cloexec} on OSes that support it.

/cc @stjepang

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -24,13 +24,16 @@ use std::time::{Duration, Instant};
 
 use libc::{self, c_void, socklen_t, ssize_t};
 
-use crate::Domain;
+use crate::{Domain, Type};
 
 #[allow(non_camel_case_types)]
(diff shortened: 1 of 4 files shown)


Thomasdezeeuw:
Seems like a sporadic CI failure, I can't reproduce it locally and the pr doesn't touch any of the code tested.

src/sys/unix.rs
```diff
@@ -94,6 +84,45 @@ impl Domain {
...
+    /// Linux, NetBSD and OpenBSD.
+    #[cfg(any(
+        target_os = "android",
+        target_os = "dragonfly",
+        target_os = "freebsd",
+        target_os = "linux",
+        target_os = "netbsd",
+        target_os = "openbsd"
+    ))]
+    pub fn non_blocking(self) -> Type {
```
Thomasdezeeuw: Btw this is one of the reasons I wanted to restructure the code.

PR #65, opened 2020-03-31
Replace cvt function with a new syscall macro

Thomasdezeeuw:


src/sys/unix.rs
```diff
@@ -345,13 +337,12 @@ impl Socket {
 
         let socket = match socket {
             Some(socket) => socket,
-            None => unsafe {
-                let fd =
-                    cvt_r(|| libc::accept(self.fd, &mut storage as *mut _ as *mut _, &mut len))?;
```
Thomasdezeeuw: This no longer reties on `ErrorKind::Interrupted`.

src/sys/unix.rs
```diff
@@ -327,15 +322,12 @@ impl Socket {
         let mut socket = None;
         #[cfg(target_os = "linux")]
         {
-            let res = cvt_r(|| unsafe {
```
Thomasdezeeuw: This no longer reties on `ErrorKind::Interrupted`.

alexcrichton:
Personally at least I prefer to avoid usage of macros wherever possible because they tend to always just add an extra layer of complication. I'll leave this up to @stjepang though

Thomasdezeeuw:
@alexcrichton I agree that macros often complicates things, but in this case I disagree. The macro is itself is fairly simple and encapsulated in its own code block, so its won't overwrite variables. Its also been in the Mio codebase for a while without problems.

Thomasdezeeuw:
@stjepang have you had any time too look at this?

Thomasdezeeuw:
@stjepang @alexcrichton any time to take a look at this?

stjepang:
Hey, sorry for the delay in reviewing this!

I personally find the `syscall!` macro nice and readable, although that perhaps has something to do with my familiarity with `mio`'s codebase. And to be honest, I've no idea what "cvt" stands for so this is an improvement to me (:

This crate is very heavy on syscalls anyway so it's not like adding a macro for such a common thing in the codebase hurts readablity. So I find the use of the macro justified.

Quick question to make sure I understand: why do we not retry on EINTR anymore? The standard library loops on EINTR, so what is the reason to deviate here?

I worry that by not handling EINTR, we push the burden of the responsibility of handling this error to the user, who might not even think about this error. So I worry this could make code more fragile and exhibit bugs in production.

Thomasdezeeuw:
> Quick question to make sure I understand: why do we not retry on EINTR anymore? The standard library loops on EINTR, so what is the reason to deviate here?

I removed it because users might want to do something themselves if `EINTR` comes up, for example when handling a signal.

> I worry that by not handling EINTR, we push the burden of the responsibility of handling this error to the user, who might not even think about this error. So I worry this could make code more fragile and exhibit bugs in production.

I agree that this puts additional burden on the user, but I do hope they think about such errors when using a low level library like socket2.

I don't really have a strong opinion on this, so I'm OK with changing it back if you think that is the better route.

stjepang:
Hmmm. Do you perhaps know, does the standard library have a consistent guideline on how it handles EINTR? Does it spin on EINTR? Does it always it return `ErrorKind::Interrupted`?

Thomasdezeeuw:
I don't know about a guideline, but from a quick look it seems it does different things for different functions. For example (on Unix) `TcpStream::connect` retries: https://github.com/rust-lang/rust/blob/fb5615a4771ea3d54256f969dc84d2dfd38d812c/src/libstd/sys_common/net.rs#L220, but for `setsockopt` (used by e.g. `TcpStream::set_ttl`) it doesn't retry: https://github.com/rust-lang/rust/blob/fb5615a4771ea3d54256f969dc84d2dfd38d812c/src/libstd/sys_common/net.rs#L61-L67, the definitions of `cvt` and `cvt_r`: https://github.com/rust-lang/rust/blob/fb5615a4771ea3d54256f969dc84d2dfd38d812c/src/libstd/sys/unix/mod.rs#L142-L157.

stjepang:
I'd personally prefer to do the same thing the standard library does, that should be cause the least amount of surprises.

I've definitely been bitten by code that works in tests but suddenly breaks in actual use because I forgot about `Interrupted` :/ (and it can be hard to find all interruptible functions in a codebase).

Thomasdezeeuw:
I've rebased this on master. @stjepang the last commit changes the calls to `connect(2)` and `accept(2)` to retry on `EINTR`, matching what libstd does. Other system calls, perhaps somewhat surprisingly including reading/writing, do not retry on `EINTR`.

I don't really think this is the right approach, I would say to not handle `EINTR` at all and just have a big warning saying its not handled by `socket2`. But on the other hand I understand wanting to match libstd.

Thomasdezeeuw:
Shall I remove the last commit? Which means that `EINTR` not handled by `socket2`. @alexcrichton you also ok with that?

stjepang:
I'm okay with that. 👍

alexcrichton:
Ok I've invited y'all to the crates.io crate and to the repo, wanna test that out w/ merging this?

Thomasdezeeuw:
I'll drop the last commit add a warning and see if I can merge it.

PR #70, opened 2020-04-07
Remove all features

Thomasdezeeuw:
This removes all features and exposes the functions previously behind a feature on all platforms that support it by default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,5 @@ redox_syscall = "0.1.38"
 tempdir = "0.3"
 
 [features]
-reuseport = []
 pair = []
 unix = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,5 @@ redox_syscall = "0.1.38"
 tempdir = "0.3"
 
 [features]
-default = []
 # Enable all API, even ones not available on all OSes.
 all = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,3 @@ redox_syscall = "0.1.38"
 
 [dev-dependencies]
 tempdir = "0.3"
-
-[features]
-unix = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,3 +28,8 @@ redox_syscall = "0.1.38"
 
 [dev-dependencies]
 tempdir = "0.3"
+
+[features]
+default = []
+# Enable all API, even ones not available on all OSes.
+all = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,5 +30,4 @@ redox_syscall = "0.1.38"
 tempdir = "0.3"
 
 [features]
-pair = []
 unix = []
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -167,6 +167,7 @@ impl Type {
     pub const DGRAM: Type = Type(sys::SOCK_DGRAM);
 
     /// Type corresponding to `SOCK_SEQPACKET`.
+    #[cfg(all(feature = "all", not(target_os = "redox")))]
     pub const SEQPACKET: Type = Type(sys::SOCK_SEQPACKET);
 
     /// Type corresponding to `SOCK_RAW`.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -171,7 +171,6 @@ impl Type {
     pub const SEQPACKET: Type = Type(sys::SOCK_SEQPACKET);
 
     /// Type corresponding to `SOCK_RAW`.
-    #[cfg(not(target_os = "redox"))]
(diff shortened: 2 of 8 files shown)


alexcrichton:
Personally I do not feel this is the best way to expose an API in a cross-platform crate. I feel there needs to be *some* level of opt-in to platform-specific features. In libstd it's extension traits, and here it's features. I don't think that the APIs should be usable on unix with zero opt-in myself.

Thomasdezeeuw:
Due to the low-level nature of the create wouldn't that create a huge amount of features? For example should additional `Type` flags be behind a feature flag? https://github.com/alexcrichton/socket2-rs/blob/b6a607fdd60ad00284fe4514795730b60e8ea721/src/sys/unix.rs#L97-L125

I see this crate as a type wrapper around libc, exposing a nicer Rusty-er API. Creating features for all the different functions that are available on some platforms but not on other would create too many features imo. Furthermore rustdoc can now clearly show that API is platform specific, for example see [`SourceFd`](https://docs.rs/mio/0.7.0/mio/unix/struct.SourceFd.html) from Mio that is only available on Unix platforms.

alexcrichton:
Yeah I don't think it's worth having a litany of features for APIs like that, and it's a bummer they're not currently gated behind a feature when they probably should be. That being said I do not personally agree with simply putting `#[cfg]` and documentation on items, and even if it works for Mio I would personally prefer to not add it here.

Thomasdezeeuw:
@alexcrichton in your opinion what should and shouldn't be behind a feature flag? A concrete example: [`accept4(2)`](http://man7.org/linux/man-pages/man2/accept4.2.html), available on most Unix systems with a notable exception being macOS. Should that be a feature flag or not?

alexcrichton:
To be clear the way I view this is that there's a set of "tier 1" platforms. That's generally Windows/mac/Linux. If functionality is available on all of them, then there's no worries about gating the API. If functionality is only available on a subset, though, then some form of opt-in should be needed so authors can indicate "I know I'm using a platform-specific API". 

Currently that's roughly done with features (but probably not completely so). It might be best to do it with a general "platform-specific" feature rather than opting into every single platform itself instead. We could also remove crate features entirely and use a module hierarchy like libstd with extension traits, either way is fine by me.

Thomasdezeeuw:
Personally I'm not the biggest fan of the extension traits, but its a route we could take.

What about a single feature to opt-in to all API, including API not available on all platforms, which is disable by default. Something like a `platform-specific` or `os-specific` feature?

Thomasdezeeuw:
Basically a `I-know-I-am-using-API-not-available-on-all-OSes` feature.

alexcrichton:
I'd be ok with that yeah. It's still not the greatest thing in the world but we don't have a ton of options available to us in general.

Thomasdezeeuw:
I've added a `all` features which enables all API, I've also created #72 to show the `cfg` attributes on docs.rs, and the pr is rebased on master.

So let the bikeshedding over the name begin. `all` is a bit plain, maybe it should be a bit "scarier" to hint to people its not something to enable lightly?

Thomasdezeeuw:
The last commit should fix the Redox build, I had to put some more functions behind the `all` flag however.

Cargo.toml
```diff
@@ -30,6 +30,6 @@ redox_syscall = "0.1.38"
 tempdir = "0.3"
 
 [features]
-reuseport = []
-pair = []
-unix = []
+default = []
```
alexcrichton, line 33: I think that `default` can be omitted here?

src/lib.rs
```diff
@@ -162,10 +167,12 @@ impl Type {
     pub const DGRAM: Type = Type(sys::SOCK_DGRAM);
 
     /// Type corresponding to `SOCK_SEQPACKET`.
+    #[cfg(all(feature = "all", not(target_os = "redox")))]
     pub const SEQPACKET: Type = Type(sys::SOCK_SEQPACKET);
 
     /// Type corresponding to `SOCK_RAW`.
     #[cfg(not(target_os = "redox"))]
```
alexcrichton, line 174: I think this line can be dropped?

src/sockaddr.rs
```diff
@@ -59,13 +59,12 @@ impl SockAddr {
...
     ///
-    /// This function is only available on Unix when the `unix` feature is
-    /// enabled.
+    /// This function is only available on Unix.
     ///
     /// # Failure
     ///
     /// Returns an error if the path is longer than `SUN_LEN`.
-    #[cfg(all(unix, feature = "unix"))]
+    #[cfg(unix)]
```
alexcrichton, line 67: Should this be behind `feature = "all"`?

Thomasdezeeuw:
@alexcrichton I've addressed all feedback.

PR #71, opened 2020-04-07
Enable some stricter warnings

Thomasdezeeuw:
Also change the examples to the try operator (`?`) instead of
unwrapping.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,22 +20,30 @@
 //! # Examples
 //!
 //! ```no_run
+//! # fn main() -> std::io::Result<()> {
 //! use std::net::SocketAddr;
 //! use socket2::{Socket, Domain, Type};
 //!
 //! // create a TCP listener bound to two addresses
-//! let socket = Socket::new(Domain::IPV6, Type::STREAM, None).unwrap();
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -20,22 +20,37 @@
...
 //! ```
 
 #![doc(html_root_url = "https://docs.rs/socket2/0.3")]
-#![deny(missing_docs)]
+#![deny(
+    missing_docs,
+    missing_debug_implementations,
+    rust_2018_idioms,
+    unused_imports,
+    dead_code
```
alexcrichton, line 47: Could this be left as `missing_docs` and maybe missing debug implementations? The others I think are best covered by `deny(warnings)`

Thomasdezeeuw: `rust_2018_idioms` and `missing_debug_implementations` are covered by warnings by default. Should I turn those into warnings and remove the rest?

Thomasdezeeuw: I've reduced it to `#![deny(missing_docs, missing_debug_implementations, rust_2018_idioms)]` which aren't enabled by default.

PR #73, opened 2020-04-27
apple: set_bound_interface

balboah:
This call works on MacOS & iOS for binding a socket to a particular interface.
It's especially useful on iOS when implementing a VPN service, where you need to decide if a new socket should be routed through the VPN or outside of it. It's not possible to do this with only the `bind(addr)`.

I'm not sure if it belongs in here or not and Linux/Windows implementation is not there yet.
What do you think?

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -368,6 +368,22 @@ impl Socket {
         self.inner.set_ttl(ttl)
     }
 
+    /// Set the value of IP_BOUND_IF option on this socket.
+    ///
+    /// Name is used for finding the correct interface index.
+    #[cfg(all(feature = "all", not(target_os = "redox")))]
+    pub fn set_bound_interface(&self, name: &str) -> io::Result<()> {
+        self.inner.set_bound_interface(name)
+    }
+
+    /// Set the value of IPV6_BOUND_IF option on this socket.
+    ///
+    /// Name is used for finding the correct interface index.
+    #[cfg(all(feature = "all", not(target_os = "redox")))]
+    pub fn set_bound_interface_v6(&self, name: &str) -> io::Result<()> {
+        self.inner.set_bound_interface_v6(name)
+    }
+
     /// Gets the value of the `IPV6_UNICAST_HOPS` option for this socket.
     ///
     /// Specifies the hop limit for ipv6 unicast packets
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -370,18 +370,18 @@ impl Socket {
 
     /// Set the value of IP_BOUND_IF option on this socket.
     ///
-    /// Name is used for finding the correct interface index.
+    /// See libc::if_nametoindex on mapping interface name to interface index.
     #[cfg(all(feature = "all", not(target_os = "redox")))]
-    pub fn set_bound_interface(&self, name: &str) -> io::Result<()> {
-        self.inner.set_bound_interface(name)
+    pub fn set_bound_interface(&self, if_index: sys::c_uint) -> io::Result<()> {
+        self.inner.set_bound_interface(if_index)
     }
 
     /// Set the value of IPV6_BOUND_IF option on this socket.
     ///
-    /// Name is used for finding the correct interface index.
+    /// See libc::if_nametoindex on mapping interface name to interface index.
     #[cfg(all(feature = "all", not(target_os = "redox")))]
-    pub fn set_bound_interface_v6(&self, name: &str) -> io::Result<()> {
-        self.inner.set_bound_interface_v6(name)
+    pub fn set_bound_interface_v6(&self, if_index: sys::c_uint) -> io::Result<()> {
+        self.inner.set_bound_interface_v6(if_index)
     }
 
     /// Gets the value of the `IPV6_UNICAST_HOPS` option for this socket.
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -546,7 +546,7 @@ impl Socket {
     #[cfg(target_vendor = "apple")]
     pub fn set_bound_interface(&self, name: &str) -> io::Result<()> {
         const IP_BOUND_IF: libc::c_int = 25;
-        let name = CString::new(name).unwrap();
+        let name = CString::new(name)?;
         unsafe {
             let index = libc::if_nametoindex(name.as_ptr());
             self.setsockopt(libc::IPPROTO_IP, IP_BOUND_IF, index)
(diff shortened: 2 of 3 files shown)


alexcrichton:
Seems plausible to me! It's ok to have the API only present on some platforms for now gated behind the `all` feature as well

balboah:
@alexcrichton ready for review, let me know if I got the hang of the "all" feature gate

src/sys/unix.rs
```diff
@@ -542,6 +543,36 @@ impl Socket {
         unsafe { self.setsockopt(libc::IPPROTO_IP, libc::IP_TTL, ttl as c_int) }
     }
 
+    #[cfg(target_vendor = "apple")]
+    pub fn set_bound_interface(&self, name: &str) -> io::Result<()> {
+        const IP_BOUND_IF: libc::c_int = 25;
+        let name = CString::new(name)?;
+        unsafe {
+            let index = libc::if_nametoindex(name.as_ptr());
```
stjepang, line 551: I think we need to check whether `index` is valid. If the index is 0, that indicates an error and we should inspect `errno` in that case.

alexcrichton: Hm ideally this'd just be one "syscall", so perhaps the argument here should be the index and not the name?

balboah: good input, I’ll split it out into two functions instead

balboah: Updated to remove the additional name to interface index lookup since it's already available in libc. I believe this resolves your comments

alexcrichton:
Since this is only available on osx, I think the `#[cfg]` should be `target_os = "macos"` and other platforms shouldn't need an implementation that panics?

balboah:
I believe that at least Linux has something similar with "SO_BINDTODEVICE". Do you want to remove the "all" feature and instead remove the non-apple placeholders?

alexcrichton:
I would personally prefer to not expose functions which panic, but instead those functions shoudl be removed, so the only way then to get this to compile would be to put it behind a more restrictive `#[cfg]`. It can be relaxed if the linux implementation comes along later.

balboah:
Sure, I might have misunderstood your earlier comment. I thought you wanted to have it available for all platforms if "all" feature is enabled, even if they're not working.

> It's ok to have the API only present on some platforms for now gated behind the all feature as well

As you can see I'm already filtering on `#[cfg(target_vendor = "apple")]`. I'll remove the other `unimplemented!()` functions

alexcrichton:
Er well, yeah, I'm basically saying that the `unimplemented!()` should all be removed and then the `#[cfg]` should be adjusted as necessary to get things compiling.

src/sys/windows.rs
```diff
@@ -432,6 +432,16 @@ impl Socket {
         unsafe { self.setsockopt(IPPROTO_IP, IP_TTL, ttl as c_int) }
     }
 
+    #[cfg(feature = "all")]
+    pub fn set_bound_interface(&self, _if_index: c_uint) -> io::Result<()> {
```
Thomasdezeeuw, line 436: These two functions should just be removed. Behind the `all` feature flag is OK to have missing functions.

src/socket.rs
```diff
@@ -368,6 +368,22 @@ impl Socket {
         self.inner.set_ttl(ttl)
     }
 
+    /// Set the value of IP_BOUND_IF option on this socket.
+    ///
+    /// See libc::if_nametoindex on mapping interface name to interface index.
+    #[cfg(all(feature = "all", not(target_os = "redox")))]
+    pub fn set_bound_interface(&self, if_index: sys::c_uint) -> io::Result<()> {
```
Thomasdezeeuw, line 375: We normally use `i32` instead of `sys::c_int`, so I think `u32` should be used here.

src/sys/unix.rs
```diff
@@ -542,6 +542,28 @@ impl Socket {
         unsafe { self.setsockopt(libc::IPPROTO_IP, libc::IP_TTL, ttl as c_int) }
     }
 
+    #[cfg(target_vendor = "apple")]
+    pub fn set_bound_interface(&self, if_index: libc::c_uint) -> io::Result<()> {
+        const IP_BOUND_IF: libc::c_int = 25;
```
Thomasdezeeuw, line 547: This should come from the `libc` crate.

src/sys/unix.rs
```diff
@@ -542,6 +542,28 @@ impl Socket {
...
+        unsafe { self.setsockopt(libc::IPPROTO_IP, IP_BOUND_IF, if_index) }
+    }
+
+    #[cfg(all(feature = "all", not(target_vendor = "apple")))]
+    pub fn set_bound_interface(&self, _if_index: libc::c_uint) -> io::Result<()> {
```
Thomasdezeeuw, line 552: Remove this function, same for `set_bound_interface_v6`.

src/sys/unix.rs
```diff
@@ -542,6 +542,28 @@ impl Socket {
         unsafe { self.setsockopt(libc::IPPROTO_IP, libc::IP_TTL, ttl as c_int) }
     }
 
+    #[cfg(target_vendor = "apple")]
```
Thomasdezeeuw, line 545: This should be `#[cfg(all(feature = "all", target_vendor = "apple"))]`.

src/socket.rs
```diff
@@ -368,6 +368,22 @@ impl Socket {
         self.inner.set_ttl(ttl)
     }
 
+    /// Set the value of IP_BOUND_IF option on this socket.
```
Thomasdezeeuw, line 371: Quotes around `IP_BOUND_IF`, same for the other function.

src/socket.rs
```diff
@@ -368,6 +368,22 @@ impl Socket {
...
+    ///
+    /// See libc::if_nametoindex on mapping interface name to interface index.
```
Thomasdezeeuw, line 373: Quotes around `libc::if_nametoindex`, same for the other function.

src/socket.rs
```diff
@@ -368,6 +368,22 @@ impl Socket {
...
+    #[cfg(all(feature = "all", not(target_os = "redox")))]
```
Thomasdezeeuw, line 374: This should be `#[cfg(all(feature = "all", target_vendor = "apple"))]` as its only available on Apple platforms from what I can gather from the pr.

balboah:
sorry to waste your time, this PR grew larger than what I had foreseen and I don't think this narrow use case is worth to spend more time on right now

Thomasdezeeuw:
@balboah sorry to hear that. If you're discouraged by my comments note that they're all just all small things.

PR #74, opened 2020-04-30
Change crate docs to reflect new crate goals

Thomasdezeeuw:
If everyone agrees with these new goals:

* Providing a minimal API around the libc/system calls.
* Expose as much of the API as possible, possibly behind the `all` flag for API not available on all the (tier 1) OSes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,15 +15,23 @@
 // is platform specific, e.g. `Domain::UNIX`, is defined in `src/sys/*.rs` and
 // only for the platforms that support it.
 
-//! Utilities for handling sockets
+//! Utilities for creating and using sockets.
 //!
-//! This crate is sort of an evolution of the `net2` crate after seeing the
-//! issues on it over time. The intention of this crate is to provide as direct
-//! as possible access to the system's functionality for sockets as possible. No
-//! extra fluff (e.g. multiple syscalls or builders) provided in this crate.
-//! This also means it doesn't handle errors such as `EINTR`. As a result using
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -8,15 +8,30 @@
...
+// only for the platforms that support it.
+
+//! Utilities for creating and using sockets.
+//!
+//! The goal of this crate is to create and use a socket using advanced
+//! configuration options (those that are not available in the types in the
+//! standard library) without using any unsafe code.
+//!
+//! This crate provides as direct as possible access to the system's
+//! functionality for sockets as possible, this means **no** cross-platform
```
stjepang, line 25:
```suggestion
//! functionality for sockets, this means **no** cross-platform
```

src/lib.rs
```diff
@@ -8,15 +8,30 @@
...
+//! utilities, no extra goodies, no creature comforts. It is up to the user to
```
stjepang, line 26: Well... we *do* get plenty of cross-platform utilities as well as platform-specific options :) But I agree with the sentiment.

src/lib.rs
```diff
@@ -8,15 +8,30 @@
...
+//! known how to use sockets when using this crate. *If you don't know how to
```
stjepang, line 27:
```suggestion
//! know how to use sockets when using this crate. *If you don't know how to
```

src/lib.rs
Thomasdezeeuw: I've changed the wording a bit. Could you take another look?

If you ok with it I'll squash the commits and merge.

PR #75, opened 2020-05-01
Refactor SockAddr

Thomasdezeeuw:
* Moves raw type definitions, such as sockaddr_storage, into the sys
  module importing them in the sockaddr modules.
* Moves SockAddr::unix to src/sys/unix.rs.
* Moves the tests into the tests module.

### DIFF
--- a/src/sockaddr.rs
+++ b/src/sockaddr.rs
@@ -55,16 +55,9 @@ impl SockAddr {
     /// Returns this address as a `SocketAddr` if it is in the `AF_INET` (IP v4)
     /// or `AF_INET6` (IP v6) family, otherwise returns `None`.
     pub fn as_std(&self) -> Option<SocketAddr> {
-        let storage: *const sockaddr_storage = (&self.storage) as *const _;
-        match self.storage.ss_family as c_int {
-            AF_INET => Some(SocketAddr::V4(unsafe {
-                *(storage as *const sockaddr_in as *const _)
-            })),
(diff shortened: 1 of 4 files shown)


src/sockaddr.rs
```diff
@@ -27,153 +16,86 @@ pub struct SockAddr {
...
+    pub fn as_ptr(&self) -> *const sockaddr {
+        &self.storage as *const _ as *const _
     }
 
-    unsafe fn as_<T>(&self, family: sa_family_t) -> Option<T> {
-        if self.storage.ss_family != family {
-            return None;
+    /// Returns this address as a `SocketAddr` if it is in the `AF_INET` (IP v4)
+    /// or `AF_INET6` (IP v6) family, otherwise returns `None`.
+    pub fn as_std(&self) -> Option<SocketAddr> {
```
stjepang, line 57: This can be a safe one-liner :)

```rust
self.as_inet().or_else(|| self.as_inet6())
```

Thomasdezeeuw: We need to call `Into::into` first, so we'll get.

```rust
        self.as_inet()
            .map(|a| a.into())
            .or_else(|| self.as_inet6().map(|a| a.into()))
```

Not quite a one-liner, but if you prefer it I'll change it.

stjepang: Up to you - it reduces unsafe code a bit though.

Thomasdezeeuw:
I've change `as_std` to `as_inet` and `as_inet6`.

PR #76, opened 2020-05-01
[WIP] Refactor of Socket

Thomasdezeeuw:
This is going to be a big one...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -54,6 +54,7 @@
 #![doc(test(attr(deny(warnings))))]
 
 use std::net::SocketAddr;
+use std::ops::BitOr;
 
 /// Macro to implement `fmt::Debug` for a type, printing the constant names
 /// rather than a number.
@@ -221,6 +222,44 @@ impl From<c_int> for Protocol {
 
 impl From<Protocol> for c_int {
     fn from(p: Protocol) -> c_int {
+        p.0
+    }
+}
+
+/// Flags used in [`Socket::send_with_flags`], used in the `send(2)` function.
+///
+/// Multiple flags can be OR-ed together.
+// TODO: add a nicer `fmt::Debug` implementation.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct SendFlag(sys::c_int);
+
+impl SendFlag {
+    /// No flags specified.
+    pub const NONE: SendFlag = SendFlag(0);
+
+    /// Flag corresponding to `MSG_DONTROUTE`.
+    pub const DONTROUTE: SendFlag = SendFlag(sys::MSG_DONTROUTE);
+
+    /// Flag corresponding to `MSG_OOB`.
+    pub const OOB: SendFlag = SendFlag(sys::MSG_OOB);
+}
+
+impl BitOr for SendFlag {
+    type Output = Self;
+
+    fn bitor(self, rhs: Self) -> Self {
+        SendFlag(self.0 | rhs.0)
+    }
+}
+
+impl From<c_int> for SendFlag {
+    fn from(p: c_int) -> SendFlag {
+        SendFlag(p)
+    }
+}
+
+impl From<SendFlag> for c_int {
+    fn from(p: SendFlag) -> c_int {
         p.0
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -263,3 +263,44 @@ impl From<SendFlag> for c_int {
         p.0
     }
 }
+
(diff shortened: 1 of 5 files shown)


Thomasdezeeuw:
This pr shows how I intend to refactor the `Socket` type. Changing `Socket` to be a wrapper around a fd/socket, rather then a `sys::Socket` type to reduce the repetition of the functions etc. With this I hope to reduce the number of lines of code. We did something similar in Mio and it removes ~1k lines of code with no change in functionally, a big win for maintenance.

I also want to add some more helper types such as `SendFlag` and `RecvFlag` to remove raw `c_int` arguments all together. This following the same pattern as for `Domain`, `Type` and `Protocol`.

Finally I want to group some of the methods on `Socket` together, e.g. all methods changing `IP_PROTO` and `IPPROTO_IPV6` level options into a single impl block. This is to provide a better overview of the available methods in the generated documentation. This is just an experiment I don't know how it will turn out.

@alexcrichton @stjepang any feedback would be welcome.

P.S. I will split this up into smaller, more reviewable prs.

oskardrums:
Hi,
If you are already looking at refactoring `Socket`, there is an issue I would like to raise: there are many combinations of socket creation parameters and following method calls which are strictly wrong. A simple example would be trying to call `accept` with a non-`SOCK_STREAM` socket. 
These "bad" combinations are a fact of the posix sockets interface, but it doesn't have to be a fact of the rust socket interface :)
I think it would be lovely if we could prevent this kind of errors (in compile time!), as well as save people the trouble of remembering which domain/type/protocol combination produces an e.g. raw IPv4 socket, presumably by creating a typed interface for different sockets kinds - each implementing only supported methods.

I found this issue quite frustrating while working with networking in rust but didn't have the time to solve it yet (I did [explore a solution](https://github.com/oskardrums/bs) that was rather specific to my own needs).

Anyways, if you would like a hand I'd be glad to help.

Thomasdezeeuw:
@oskardrums I agree with your idea, but I'm not sure socket2 is the library for that. The standard library and high-level libraries such as Mio or Tokio already provide such types.

I think socket2 such aim more advanced configuration, which requires knowledge on using sockets on the user's part.

alexcrichton:
I don't really personally have the time to review this myself, and at this point with this it's basically not really my library any more. I'm not really sure what to do about that myself. If this differs significantly from the original design then I don't think it belongs in this crate, but if it's basically the same thing but with a different veneer then it seems fine to add here. I don't really have the time to figure this out from the thousands of lines of changes here though.

oskardrums:
> 
> 
> @oskardrums I agree with your idea, but I'm not sure socket2 is the library for that. The standard library and high-level libraries such as Mio or Tokio already provide such types.
> 
> I think socket2 such aim more advanced configuration, which requires knowledge on using sockets on the user's part.

@Thomasdezeeuw do you think mio would be a better fit for e.g. `RawSocket`, `SctpSocket` (perhaps as an extension crate)?
I get that `socket2` may not be the best place for such specifically typed sockets, especially reading Alex's last comment :)

Thomasdezeeuw:
> I don't really personally have the time to review this myself, and at this point with this it's basically not really my library any more. I'm not really sure what to do about that myself. If this differs significantly from the original design then I don't think it belongs in this crate, but if it's basically the same thing but with a different veneer then it seems fine to add here. I don't really have the time to figure this out from the thousands of lines of changes here though.

I get it. Like I said it I will split this it over multiple prs, I just wanted to my (current) ideas for the crates outlined in https://github.com/alexcrichton/socket2-rs/pull/76#issuecomment-622881708. I would say the ideas are inline with the original design, but stripping a layer from it (removing the `sys::Socket` type).

Thomasdezeeuw:
> @Thomasdezeeuw do you think mio would be a better fit for e.g. `RawSocket`, `SctpSocket` (perhaps as an extension crate)?
> I get that `socket2` may not be the best place for such specifically typed sockets, especially reading Alex's last comment :)

Mio already has it to an extend: `TcpStream`, `TcpListener`, `UdpSocket` and the UDS types are already typed-sockets (there all wrappers around file descriptors on Unix). But right now Mio is sticking to the same API std lib provides. I would say another crate might be for now to experiment with, once it a bit more stable it might be worth adding to Mio or even std lib.

Thomasdezeeuw:
This refactor is complete.

For anyone keeping track: the lines of code when down from 2967 to 2450 (according to tokei), while we've added many features such as `accept4(2)`, a great TCP keepalive API and things such as `Type::nonblocking`.

PR #77, opened 2020-05-02
Haiku: no SOCK_RDM support in Haiku

nielx:
Do not add SOCK_RDM when compiling for Haiku.


stjepang:
I don't know anything about Haiku either. Also okay with merging this PR, but it would be a good idea to do `cargo check` for haiku (and other platforms) in CI.

Thomasdezeeuw:
I'm going to merge this, but I agree with @stjepang we should add more CI check tests so I've opened: #78.

Thomasdezeeuw:
Thanks @nielx

PR #82, opened 2020-05-23
Add as_unix method to SockAddr + Fix Windows Clippy check

PerfectLaugh:
Have been tested on smol tests.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -12,6 +12,8 @@ use std::net::{self, Ipv4Addr, Ipv6Addr};
 #[cfg(feature = "all")]
 use std::os::unix::ffi::OsStrExt;
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd};
+#[cfg(feature = "all")]
+use std::os::unix::net::SocketAddr;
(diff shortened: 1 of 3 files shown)


src/sys/unix.rs
```diff
@@ -236,6 +238,43 @@ impl SockAddr {
...
+        };
+
+        #[allow(dead_code)]
+        struct SocketAddrInternal {
+            addr: libc::sockaddr_un,
+            len: libc::socklen_t,
+        }
+
+        Ok(unsafe {
+            std::mem::transmute(SocketAddrInternal {
```
Thomasdezeeuw, line 272: The assumption that `SocketAddrInternal` is the correct layout for `std::os::unix::net::SocketAddr` is invalid. We've just fixed this for IP socket addresses, so let's not regress for Unix.

Thomasdezeeuw (changes requested):
This would need to some kind of constructor in `std::os::unix::net::SocketAddr`.

Thomasdezeeuw:
Since we still don't have a correct way to create `std::os::unix::net::SocketAddr` I'm closing this issue.

PR #84, opened 2020-06-02
fix: unify sigpipe settings for macos and ios

### LINKED ISSUE #83: SO_NOSIGPIPE not set on ios
I am experiencing trouble with sigpipes using socket2 on ios. The code already sets nosigpipe by default for macos, is there a reason this is not done for ios as well?

dignifiedquire:
Closes #83


Thomasdezeeuw:
Redox fails but that is unrelated, so I'm merging. Thanks @dignifiedquire

PR #88, opened 2020-06-25
Correc byte ordering to convert numbers to Ipv4 addresses

bltavares:
This commit changes the byte order when converting IPv4 Addresses, as
the order was reversed.

The test program used to validate it works was sending a multicast
address to `join_multicast_v4` which was not accepted by either Windows
or Linux. Now the test program compiles and run without issues.

```rust
use socket2::{Domain, Protocol, Socket, Type};
use std::net::Ipv4Addr;

fn main() {
    let socket = Socket::new(Domain::IPV4, Type::DGRAM, Some(Protocol::UDP))
        .expect("could not create socket");
    socket
        .join_multicast_v4(&Ipv4Addr::new(224, 0, 0, 251), &Ipv4Addr::UNSPECIFIED)
        .expect("could not join multicast group");
}
```

Fixes https://github.com/alexcrichton/socket2-rs/issues/87

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -1141,11 +1141,11 @@ fn timeval2dur(raw: libc::timeval) -> Option<Duration> {
 
 fn to_s_addr(addr: &Ipv4Addr) -> libc::in_addr_t {
     let octets = addr.octets();
-    u32::from_ne_bytes(octets).to_be()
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
Thanks @bltavares.

This does show we need additional testing.

PR #92, opened 2020-08-13
Fix mutability of buffers for a few send_* functions

coolreader18:


coolreader18:
No problem! Would you be able to publish a new release to crates.io?

stjepang:
@Thomasdezeeuw Just to double-check because I haven't followed the latest changes very closely, do you think v0.3.13 can be published as is?

Thomasdezeeuw:
I think there some more issues like this, but it didn't get any worse on master. I'm ok with a release to fix this (and some other small issues).

stjepang:
Published v0.3.13

Thomasdezeeuw:
@stjepang sorry about the v0.3.13/14 releases, I took a couple of days off (perfectly timed of course) and I forgot about the breaking changes. Maybe we should create a v0.3.x branch and make master the new development branch targeting v0.4.0.

PR #102, opened 2020-09-20
Added support for setting the socket mark option (SO_MARK)

ammarzuberi:


src/socket.rs
```diff
@@ -370,6 +370,19 @@ impl Socket {
...
+    /// Sets the value for the `SO_MARK` option on this socket.
+    ///
+    /// This value sets the socket mark field for each packet sent through
+    /// this socket. Changing the mark can be used for mark-based routing
+    /// without netfilter or for packet filtering.
+    ///
+    /// This function is only available on Unix and requires the
+    /// `CAP_NET_ADMIN` capability.
+    #[cfg(target_os = "linux")]
+    pub fn set_mark(&self, mark: u32) -> io::Result<()> {
```
Thomasdezeeuw, line 382: I think `mark` should be `libc::c_int`, then we don't have to convert it in `sys/unix.rs`.

Thomasdezeeuw (changes requested):
Small nit, but LGTM otherwise.

@ammarzuberi do note that we use the master branch for development of the v0.4.x release. You might want to create another pr against the v0.3.x branch.

src/socket.rs
ammarzuberi: Would that not push the cast into the application code? Using `c_int` is consistent with `set_ttl` etc.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @ammarzuberi.

PR #106, opened 2020-10-01
Use SOCK_CLOEXEC flag on more platforms, and for socketpair.

de-vri-es:
This is a bit of a follow-up to #105, but I'm less sure about the error path so I though it prudent to split into two PRs.

The PR does two things:
* Use the `SOCK_CLOEXEC` flag on more platforms.
* Use the `SOCK_CLOEXEC` flag for socketpair creation.

All the recent versions of those platforms support the `SOCK_CLOEXEC` flag. The one thing I'm not sure of is what the error code is that will be returned by older versions of those platforms.

I would guess for this one (all the BSD man pages say something similar):
> [EPROTOTYPE] The combination of the specified protocol and type is not supported.

Since it's a flag orred into the `type` parameter, this seems like the error that should trigger.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -132,13 +132,14 @@ impl Type {
     /// # Notes
     ///
     /// This function is only available on Android, DragonFlyBSD, FreeBSD,
-    /// Linux, NetBSD and OpenBSD.
+    /// Illumos, Linux, NetBSD and OpenBSD.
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw (approved):

de-vri-es:
This no longer applies. I will submit a new PR soon.

PR #107, opened 2020-10-01
Implement vectored I/O.

de-vri-es:
This PR implements vectored I/O.

It adds 4 functions, all of which take a flags argument.
* `recv_vectored`
* `recv_from_vectored`
* `send_vectored`
* `send_to_vectored`

I also considered adding separate `_with_flags` versions, and `peek_vectored` and `peek_from_vectored`, but that would have been a total of 12 new functions. To keep the API from exploding I figured this would be better: you get full flexibility, and vectored I/O is probably advanced enough that the mandatory flags argument is acceptable.

Atleast, that was my conclusion.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -6,6 +6,8 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+#[cfg(not(target_os = "redox"))]
+use std::io::{IoSlice, IoSliceMut};
 use std::io::{Read, Write};
 use std::net::Shutdown;
 use std::net::{self, Ipv4Addr, Ipv6Addr};
@@ -78,6 +80,20 @@ macro_rules! syscall {
         }
     }};
 }
+
+#[cfg(any(target_os = "android", all(target_os = "linux", target_env = "gnu")))]
+type IovLen = usize;
+
+#[cfg(any(
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "ios",
+    all(target_os = "linux", target_env = "musl",),
+    target_os = "macos",
+    target_os = "netbsd",
+    target_os = "openbsd",
+))]
+type IovLen = c_int;
 
 /// Unix only API.
 impl Domain {
@@ -488,6 +504,51 @@ impl Socket {
         Ok((n as usize, addr))
     }
 
+    #[cfg(not(target_os = "redox"))]
+    pub fn recv_vectored(
+        &self,
+        bufs: &mut [IoSliceMut<'_>],
+        flags: c_int,
+    ) -> io::Result<(usize, bool)> {
+        let mut msg = libc::msghdr {
+            msg_name: std::ptr::null_mut(),
+            msg_namelen: 0,
+            msg_iov: bufs.as_mut_ptr().cast(),
+            msg_iovlen: bufs.len().min(IovLen::MAX as usize) as IovLen,
+            msg_control: std::ptr::null_mut(),
+            msg_controllen: 0,
+            msg_flags: 0,
+        };
+
+        let n = syscall!(recvmsg(self.fd, &mut msg as *mut _, flags))?;
+        let truncated = msg.msg_flags & libc::MSG_TRUNC != 0;
+        Ok((n as usize, truncated))
+    }
+
+    #[cfg(not(target_os = "redox"))]
+    pub fn recv_from_vectored(
+        &self,
+        bufs: &mut [IoSliceMut<'_>],
+        flags: c_int,
+    ) -> io::Result<(usize, bool, SockAddr)> {
+        let mut storage: libc::sockaddr_storage = unsafe { mem::zeroed() };
+        let mut msg = libc::msghdr {
+            msg_name: &mut storage as *mut libc::sockaddr_storage as *mut c_void,
+            msg_namelen: mem::size_of_val(&storage) as socklen_t,
+            msg_iov: bufs.as_mut_ptr().cast(),
+            msg_iovlen: bufs.len().min(IovLen::MAX as usize) as IovLen,
+            msg_control: std::ptr::null_mut(),
+            msg_controllen: 0,
+            msg_flags: 0,
+        };
+
+        let n = syscall!(recvmsg(self.fd, &mut msg as *mut _, flags))?;
+        let truncated = msg.msg_flags & libc::MSG_TRUNC != 0;
+        let addr =
+            unsafe { SockAddr::from_raw_parts(&storage as *const _ as *const _, msg.msg_namelen) };
+        Ok((n as usize, truncated, addr))
+    }
+
     pub fn send(&self, buf: &[u8], flags: c_int) -> io::Result<usize> {
         let n = syscall!(send(
             self.fd,
@@ -507,6 +568,43 @@ impl Socket {
             addr.as_ptr(),
             addr.len(),
         ))?;
+        Ok(n as usize)
+    }
+
+    #[cfg(not(target_os = "redox"))]
+    pub fn send_vectored(&self, bufs: &[IoSlice<'_>], flags: c_int) -> io::Result<usize> {
+        let mut msg = libc::msghdr {
+            msg_name: std::ptr::null_mut(),
+            msg_namelen: 0,
+            msg_iov: bufs.as_ptr() as *mut libc::iovec,
+            msg_iovlen: bufs.len().min(IovLen::MAX as usize) as IovLen,
+            msg_control: std::ptr::null_mut(),
+            msg_controllen: 0,
+            msg_flags: 0,
+        };
+
+        let n = syscall!(sendmsg(self.fd, &mut msg as *mut libc::msghdr, flags))?;
+        Ok(n as usize)
+    }
+
+    #[cfg(not(target_os = "redox"))]
+    pub fn send_to_vectored(
+        &self,
+        bufs: &[IoSlice<'_>],
+        flags: c_int,
+        addr: &SockAddr,
+    ) -> io::Result<usize> {
+        let mut msg = libc::msghdr {
+            msg_name: addr.as_ptr() as *mut c_void,
+            msg_namelen: addr.len(),
+            msg_iov: bufs.as_ptr() as *mut libc::iovec,
+            msg_iovlen: bufs.len().min(IovLen::MAX as usize) as IovLen,
+            msg_control: std::ptr::null_mut(),
+            msg_controllen: 0,
+            msg_flags: 0,
+        };
+
+        let n = syscall!(sendmsg(self.fd, &mut msg as *mut libc::msghdr, flags))?;
         Ok(n as usize)
     }
 
--- a/src/sys/windows.rs
+++ b/src/sys/windows.rs
@@ -9,7 +9,7 @@
 use std::cmp;
 use std::fmt;
 use std::io;
-use std::io::{Read, Write};
+use std::io::{IoSlice, IoSliceMut, Read, Write};
 use std::mem;
 use std::net::Shutdown;
 use std::net::{self, Ipv4Addr, Ipv6Addr};
@@ -338,6 +338,77 @@ impl Socket {
         }
     }
 
+    pub fn recv_vectored(
+        &self,
+        bufs: &mut [IoSliceMut<'_>],
+        flags: c_int,
+    ) -> io::Result<(usize, bool)> {
+        let mut nread = 0;
+        let mut flags = flags as DWORD;
+        let ret = unsafe {
+            sock::WSARecv(
+                self.socket,
+                bufs.as_mut_ptr() as *mut WSABUF,
+                bufs.len().min(DWORD::MAX as usize) as DWORD,
+                &mut nread,
+                &mut flags,
+                ptr::null_mut(),
+                None,
+            )
+        };
+
+        let nread = nread as usize;
+        if ret == 0 {
+            Ok((nread, false))
+        } else {
+            let error = last_error();
+            match error.raw_os_error() {
+                Some(sock::WSAESHUTDOWN) => Ok((0, false)),
+                Some(sock::WSAEMSGSIZE) => Ok((nread, true)),
+                _ => Err(error),
+            }
+        }
+    }
+
+    pub fn recv_from_vectored(
+        &self,
+        bufs: &mut [IoSliceMut<'_>],
+        flags: c_int,
+    ) -> io::Result<(usize, bool, SockAddr)> {
+        let mut nread = 0;
+        let mut flags = flags as DWORD;
+        let mut storage: SOCKADDR_STORAGE = unsafe { mem::zeroed() };
+        let mut addrlen = mem::size_of_val(&storage) as c_int;
+        let ret = unsafe {
+            sock::WSARecvFrom(
+                self.socket,
+                bufs.as_mut_ptr() as *mut WSABUF,
+                bufs.len().min(DWORD::MAX as usize) as DWORD,
+                &mut nread,
+                &mut flags,
+                &mut storage as *mut SOCKADDR_STORAGE as *mut SOCKADDR,
+                &mut addrlen,
+                ptr::null_mut(),
+                None,
+            )
+        };
+
+        let truncated;
+        if ret == 0 {
+            truncated = false;
+        } else {
+            let error = last_error();
+            if error.raw_os_error() == Some(sock::WSAEMSGSIZE) {
+                truncated = true;
+            } else {
+                return Err(error);
+            }
+        }
+
+        let addr = unsafe { SockAddr::from_raw_parts(&storage as *const _ as *const _, addrlen) };
+        Ok((nread as usize, truncated, addr))
+    }
+
     pub fn send(&self, buf: &[u8], flags: c_int) -> io::Result<usize> {
         unsafe {
             let n = {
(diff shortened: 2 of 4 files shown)


src/sys/unix.rs
```diff
@@ -79,6 +79,38 @@ macro_rules! syscall {
...
 
+#[cfg(any(
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "ios",
+    target_os = "macos",
+    target_os = "netbsd",
+    target_os = "openbsd",
+))]
+const fn max_iov() -> usize {
```
Thomasdezeeuw, line 90: Might as well make these constants. I was too lazy to do that in libstd when I added `IOV_MAX` and `UIO_MAXIOV` to libc.

src/sys/unix.rs
```diff
@@ -538,6 +570,35 @@ impl Socket {
         Ok((n as usize, addr))
     }
 
+    #[cfg(not(target_os = "redox"))]
+    pub fn recv_vectored(&self, bufs: &mut [IoSliceMut<'_>], flags: c_int) -> io::Result<usize> {
+        let mut msg: libc::msghdr = unsafe { std::mem::zeroed() };
```
Thomasdezeeuw, line 575: Nit: to avoid the `unsafe` block I think you can just create `libc::msghdr` properly.

src/sys/unix.rs
```diff
@@ -538,6 +570,35 @@ impl Socket {
...
+    }
+
+    #[cfg(not(target_os = "redox"))]
+    pub fn recv_from_vectored(
+        &self,
+        bufs: &mut [IoSliceMut<'_>],
+        flags: c_int,
+    ) -> io::Result<(usize, SockAddr)> {
+        let mut storage: libc::sockaddr_storage = unsafe { mem::zeroed() };
+        let mut msg: libc::msghdr = unsafe { std::mem::zeroed() };
```
Thomasdezeeuw, line 590: Nit: same as above for the `libc::msghdr`.

src/sys/unix.rs
```diff
@@ -560,6 +621,35 @@ impl Socket {
         Ok(n as usize)
     }
 
+    #[cfg(not(target_os = "redox"))]
+    pub fn send_vectored(&self, bufs: &[IoSlice<'_>], flags: c_int) -> io::Result<usize> {
+        let mut msg: libc::msghdr = unsafe { std::mem::zeroed() };
```
Thomasdezeeuw, line 626: Nit: same as above for the `libc::msghdr`.

src/sys/unix.rs
```diff
@@ -560,6 +621,35 @@ impl Socket {
...
+    }
+
+    #[cfg(not(target_os = "redox"))]
+    pub fn send_to_vectored(
+        &self,
+        bufs: &[IoSlice<'_>],
+        flags: c_int,
+        addr: &SockAddr,
+    ) -> io::Result<usize> {
+        let mut msg: libc::msghdr = unsafe { std::mem::zeroed() };
```
Thomasdezeeuw, line 642: Nit: same as above for the `libc::msghdr`.

src/sys/windows.rs
```diff
@@ -386,6 +386,64 @@ impl Socket {
         }
     }
 
+    pub fn recv_vectored(&self, bufs: &mut [IoSliceMut<'_>], flags: c_int) -> io::Result<usize> {
+        unsafe {
```
Thomasdezeeuw, line 390: Nit: I'm not the biggest fan of these large unsafe blocks, makes it too easy to something unsafe on accident you didn't mean to.

src/sys/windows.rs
```diff
@@ -386,6 +386,64 @@ impl Socket {
...
+                buffer_count,
+                &mut nread,
+                &mut flags,
+                ptr::null_mut(),
+                None,
+            );
+
+            match ret {
+                0 => Ok(nread as usize),
+                _ if sock::WSAGetLastError() == sock::WSAESHUTDOWN => Ok(0),
```
Thomasdezeeuw, line 407: Why is the this error being ignored? Sending on a socket that is shutdown should be an error.

src/sys/windows.rs
```diff
@@ -386,6 +386,64 @@ impl Socket {
...
+                &mut flags,
+                &mut storage as *mut SOCKADDR_STORAGE as *mut SOCKADDR,
+                &mut addrlen,
+                ptr::null_mut(),
+                None,
+            );
+
+            let nread = match ret {
+                0 => nread as usize,
+                _ if sock::WSAGetLastError() == sock::WSAESHUTDOWN => 0,
```
Thomasdezeeuw, line 439: Same as above: why ignored?

Thomasdezeeuw (approved):
Couple small things and some tests would be nice, but other LGTM.

src/sys/windows.rs
de-vri-es: Yes, but when reading on a socket, unix platforms report this error as a zero-sized read. This is also done in the other `recv` functions.

src/sys/unix.rs
de-vri-es: Done :)

src/sys/unix.rs
de-vri-es: Done :)

src/sys/windows.rs
de-vri-es: I copied the style from the surrounding functions, but I do agree. I reduced the scope of the unsafe blocks in the new functions.

de-vri-es:
> Couple small things and some tests would be nice, but other LGTM.

I processed most of the feedback. They're still separate commits to aid the review, but they can be squashed before merge.

As for tests, I agree that would be good. I couldn't find any existing tests that create connected or bound sockets. Socket pairs would be perfect, but that isn't available on windows. I could use TCP or UDP sockets with localhost. Should I go for that?

src/sys/unix.rs
de-vri-es: Done :)

src/sys/unix.rs
de-vri-es: Done :)

src/sys/unix.rs
de-vri-es: Done :)

de-vri-es:
I added tests using ephemeral UDP sockets, which should be portable to all platforms, and self contained. Using datagram sockets adds the nicety that reads and writes can never be split, so that is one less thing to worry about in the tests.

I also removed the capping of the buffer count. That would be nice behaviour for stream sockets, but for datagram sockets it is invalid to split send calls.

For receive calls, you always have the possibility to get truncated datagrams, but getting a truncated message without filling all buffers is not very nice. I think an error about the buffer count is better in that case (even though the full message might have fit in the capped buffer list).

de-vri-es:
Ah, one more open question I think: what should we do about truncated messages? Ignore, return an error, or add a boolean to the return value? Or maybe a strongly typed `Truncated` enum instead of a boolean?

src/sys/unix.rs
```diff
@@ -6,7 +6,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::io::{Read, Write};
+use std::io::{IoSlice, IoSliceMut, Read, Write};
```
Thomasdezeeuw, line 9: Wouldn't this need `#[cfg(not(target_os = "redox"))]`?

src/sys/unix.rs
```diff
@@ -1198,6 +1304,22 @@ fn dur2linger(dur: Option<Duration>) -> libc::linger {
     }
 }
 
+/// Check that the number of buffers does not exceed the maximum for vectored IO.
+///
+/// For stream sockets, the buffer count could just be capped to the maximum.
+/// However, for datagram sockets it is not possible to split reads or writes,
+/// so we must return an error.
+fn check_iov_count(count: usize) -> io::Result<()> {
```
Thomasdezeeuw, line 1312: Let's not return an error, but instead use `msg_iovlen: min(bufs.len(), IOV_MAX) as c_int` in `msghdr`. All the functions return the amount of bytes read/written and partial reads/writes are something callers have to deal with anyway.

src/sys/windows.rs
```diff
@@ -1001,6 +1108,22 @@ fn dur2linger(dur: Option<Duration>) -> sock::linger {
     }
 }
 
+/// Check that the number of buffers does not exceed the maximum for vectored IO.
+///
+/// For stream sockets, the buffer count could just be capped to the maximum.
+/// However, for datagram sockets it is not possible to split reads or writes,
+/// so we must return an error.
+fn check_buffer_count(count: usize) -> io::Result<()> {
```
Thomasdezeeuw, line 1116: Same as the Unix version.

Thomasdezeeuw (commented):
Re: https://github.com/alexcrichton/socket2-rs/pull/107#discussion_r498786400, checking `WSAESHUTDOWN` on Windows.

I'm not sure all Unixes respond with `Ok(0)`, I think this is different per OS. For now I think its best to return whatever the OS returns.


> Ah, one more open question I think: what should we do about truncated messages? Ignore, return an error, or add a boolean to the return value? Or maybe a strongly typed `Truncated` enum instead of a boolean?

We should truncate the message, that is also done in lib std, so to me that seems the least surprising.

@stjepang thoughts about the points above?

src/sys/unix.rs
de-vri-es: The problem with that is that for datagram sockets partial reads and writes aren't possible. You can read truncated messages, but by then the rest of the datagram is gone.

If we silently discard buffers, you could get a truncated datagram without filling the all the buffers. I don't see a sane way for user code to deal with that. As long as the whole buffer is used you can atleast check if the full buffer capacity was used.

de-vri-es:
> Re: [#107 (comment)](https://github.com/alexcrichton/socket2-rs/pull/107#discussion_r498786400), checking `WSAESHUTDOWN` on Windows.
> 
> I'm not sure all Unixes respond with `Ok(0)`, I think this is different per OS. For now I think its best to return whatever the OS returns.

It's what posix says, so I expect all unixes to do it. Also, the other windows `recv` functions already do the same thing:

https://github.com/alexcrichton/socket2-rs/blob/c8a2f690fe83104de8cc6a5c5f2d97a4fda70456/src/sys/windows.rs#L335


> 
> > Ah, one more open question I think: what should we do about truncated messages? Ignore, return an error, or add a boolean to the return value? Or maybe a strongly typed `Truncated` enum instead of a boolean?
> 
> We should truncate the message, that is also done in lib std, so to me that seems the least surprising.

This would indeed be in line with the standard library and the existing `recv` functions.

However, one additional reason for people to use `recvmsg` instead of plain `recv` is that it can tell you a received datagrams is truncated. With `recv`, the OS doesn't even tell you. the same is true for winsock's `WSARecv` versus `recv`. It feels a bit wrong to wrap this syscall and still throw away that information.

src/sys/unix.rs
de-vri-es: Good catch, updated :)

src/sys/unix.rs
Thomasdezeeuw: Unfortunately that is problem in the `recv` system call, one I don't think we should be fixing in socket2. Users using datagram socket should know this and use an appropriate sized buffer.

de-vri-es: I agree we don't have to fix limitations of the underlying syscalls, but I also think we should not make it worse. And giving truncated messages without using the full buffers does seem worse to me.

It would also deviating from the the syscall itself does. If the buffers exceed `IOV_MAX`, the syscall fails. Ideally, I would like to pass the `bufs.len()` unchanged to the syscall and let it report the error. But this is somewhat complicated by the fact that the `msg_iovlen` field is not a `usize` but a `socklen_t` on most platforms (except GNU/Linux and Android). So we can't pass the `usize` as is.

We could compromise and only manually return an error if `bufs.len()` exceeds the capacity of the data type of `socklen_t`, and let the OS decide how to handle it in other cases.

Thomasdezeeuw: Maybe we should remove the check entirely leaving it up to the user to check it. This is a low-level crate after all.

As for the truncation: it's only a problem with receiving from an UDP socket. But users of UDP should know to use buffer of ~65k bytes to ensure the entire packet can be read.

de-vri-es: Yeah, that is probably the most pragmatic approach.

It does leave the problem that a `usize` doesn't always fit in a `socklen_t`. I'd prefer to still return an error in that case. On GNU/Linux and Android that check can be removed, since `msg_iovlen` is a `usize`.

I admit, truncating to `socklen_t::MAX` also wouldn't be the end of the world, because it seems very unlikely that anyone would ever have more than 2^32-1 buffers. But that argument kinda goes both ways :p

Thomasdezeeuw: Let's just truncate to `socklen_t`, it's the simplest solution. And maybe add a note in function docs.

de-vri-es: Changed to code now to cap to the max for the data type of `msg_iovlen`, and let the OS handle the rest.

Also, I double-checked with `libc`, it's actually `c_int`, not `socklen_t` for the BSDs and musl. (My bad.)

src/sys/windows.rs
de-vri-es: Update to just cap to `DWORD::MAX`.

de-vri-es:
Regarding truncated datagrams: there is an open PR for the standard library that adds vectored I/O (with ancillary data, very cool) to datagram sockets. That function does return the `truncated` flag. See rust-lang/rust#69864 and specifically [this function](https://github.com/rust-lang/rust/blob/9277351ff5f71f13e33d5ba84470fb1d48ac9d6c/library/std/src/sys/unix/ext/net/datagram.rs#L376-L380).

So far the only examples of vectored I/O in the standard library was for `Read`/`Write` streams, including stream sockets. There is no vectored I/O for datagram sockets yet. But if that PR is any indication, vectored I/O for datagram sockets *will* return the `truncated` flag.

Considering this, I believe it best to copy that style and also return the `truncated` flag.

Thomasdezeeuw:
@de-vri-es say we do add a `truncated` return value. What value does it provide? Suppose it returns `true` for a message I read. What do I do with the packet then? Do I just drop the packet? Do I attempt to handle the part I did read? I can't reread the packet because it's already discarded by the OS (unless I used `MSG_PEEK`/`peek`). I just don't think the flag provides any useful information (at the time you get it).

For UDP I've always used a simple rule/guideline: read the entire packet into a single 2^16 bytes buffer. Then it can't get truncated (2^16 is the maximum size of a UDP packet including the packet metadata) and you don't have to worry about it be truncated.

de-vri-es:
> @de-vri-es say we do add a `truncated` return value. What value does it provide? Suppose it returns `true` for a message I read. What do I do with the packet then? Do I just drop the packet? Do I attempt to handle the part I did read? I can't reread the packet because it's already discarded by the OS (unless I used `MSG_PEEK`/`peek`). I just don't think the flag provides any useful information (at the time you get it).

You can report an error. That is much better than acting on truncated data without knowing it.

Or you can use `MSG_PEEK | MSG_TRUNC` to get the actual size of the datagram.

The main thing is, this is a low level wrapper around the socket API, so throwing away important information doesn't make sense to me. If you don't care about the truncated flag you can ignore it, but if you do care you can't get it back if we throw it away.

> For UDP I've always used a simple rule/guideline: read the entire packet into a single 2^16 bytes buffer. Then it can't get truncated (2^16 is the maximum size of a UDP packet including the packet metadata) and you don't have to worry about it be truncated.

That's one way to deal with it, but there are more protocols than UDP. The maximum size of unix datagrams depends on kernel settings for example.

Thomasdezeeuw (approved):
I'm OK with merging this as is.

For the future I'm wondering if we should create a wrapper around `msghdr` and pass that to the function, that would allow the user to check all flags not just `MSG_CTRUNC`.

Thomasdezeeuw:
@de-vri-es could you fix the CI problems?

@stjepang you OK with the proposed API?

de-vri-es:
> For the future I'm wondering if we should create a wrapper around `msghdr` and pass that to the function, that would allow the user to check all flags not just `MSG_CTRUNC`.

A wrapper around `msghdr` is an interesting idea. It is a bit of a weird struct though, with some fields that are both input and output. It would be hard to use portably across unix and windows, but it would be useful for situations where you need complete control over the syscall.

> @de-vri-es could you fix the CI problems?

Fixed :)

Feel free to squash into a single commit when merging. That would remove some unnecessary clutter from the history. I did have to merge master into the branch to get CI to pass, so if needed I can do a manual rebase/squash when the reviews are done.

stjepang (approved):

Thomasdezeeuw:
@de-vri-es GitHub won't allow me to merge the branch as, can you squash/rebase the commits? Then I can merge.

Thomasdezeeuw:
> A wrapper around `msghdr` is an interesting idea. It is a bit of a weird struct though, with some fields that are both input and output.

To be fair so is the buffer, but I get your point.

> It would be hard to use portably across unix and windows, but it would be useful for situations where you need complete control over the syscall.

Maybe just a `SendFlag`/`RecvFlag` can be returned instead of the truncated boolean with portable `is_truncated` like functions. I've opened #113.

de-vri-es:
Squashed and rebased :)

Thomasdezeeuw:
Thanks @de-vri-es, of dank u wel :)

de-vri-es:
Haha, graag gedaan :) And thank you for the thorough review.

PR #109, opened 2020-10-07
Refactor part 1

Thomasdezeeuw:
In this rewrite the sys module will only expose the type used by the OS
to represent a socket (e.g. SOCKET on Windows or a file descriptor on
Unix) and the required functions.

This should reduce the number of lines of code for this crate. Something
similar was done in the Mio crate which reduced the SLOC by roughly 2k.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "socket2"
-version = "0.3.14"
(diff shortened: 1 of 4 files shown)


stjepang (approved):
Seems like a good plan. 👍

PR #110, opened 2020-10-08
Refactor part 2: change Socket::new to only call socket(2)

Thomasdezeeuw:
If socket2 is going to be a fully configuration socket(2) wrapper, then
the user needs to determine if things like setting CLOEXEC is desirable.
The new example shows the common case of setting all those flags.

We should consider adding a `Socket::new_with_common_flags()`, which would create a socket with all advisable flags set such as in the new example.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,7 +35,7 @@
 //!
 //! ```no_run
 //! # fn main() -> std::io::Result<()> {
-//! use std::net::SocketAddr;
+//! use std::net::{SocketAddr, TcpListener};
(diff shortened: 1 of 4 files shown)


stjepang (approved):
I agree it would be good to have a convenient constructor that sets all the advisable flags. Right now a lot of `#[cfg(...)]` attributes are needed to set them up properly on every platform. But that can be done later...

Thomasdezeeuw:
@ stjepang created #111 for such a function.

PR #112, opened 2020-10-08
Refactor part 3: add Socket::accept4 and change Socket::accept to be a call to accept(2)

Thomasdezeeuw:


src/sys/windows.rs
```diff
@@ -61,6 +61,18 @@ pub(crate) use winapi::shared::ws2def::{
 pub(crate) use winapi::shared::ws2ipdef::SOCKADDR_IN6_LH as sockaddr_in6;
 pub(crate) use winapi::um::ws2tcpip::socklen_t;
 
+/// Helper macro to execute a system call that returns an `io::Result`.
+macro_rules! syscall {
+    ($fn: ident ( $($arg: expr),* $(,)* ), $err_test: path, $err_value: expr) => {{
+        let res = unsafe { sock::$fn($($arg, )*) };
+        if $err_test(&res, &$err_value) {
+            Err(io::Error::from_raw_os_error(unsafe { sock::WSAGetLastError() }))
```
stjepang, line 69: I'm not very familiar with windows, but noticed that some code in `std` uses `WSAGetLastError()` and some uses `io::Error::last_error()`. Not sure if those are always equivalent. Do you perhaps know? If they aren't always equivalent, that is something to be mindful of because in that case, `syscall!` only works with the WSA* class of syscalls.

stjepang (approved):

src/sys/windows.rs
Thomasdezeeuw: std lib also uses `WSAGetLastError` in the net module: https://github.com/rust-lang/rust/blob/ac48e62db85e6db4bbe026490381ab205f4a614d/library/std/src/sys/windows/net.rs#L49-L51.
`io::Error::last_error` uses `GetLastError`: https://github.com/rust-lang/rust/blob/45f638bd8620592b7c50e9b483c2aa8ef9715f5d/library/std/src/sys/windows/os.rs#L22-L24

According to https://stackoverflow.com/questions/15586224/is-wsagetlasterror-just-an-alias-for-getlasterror, `WSAGetLastError` is a wrapper for `GetLastError`. 

https://docs.microsoft.com/en-us/windows/win32/winsock/error-codes-errno-h-errno-and-wsagetlasterror-2 says 
> In Winsock applications, error codes are retrieved using the WSAGetLastError function, the Windows Sockets substitute for the Windows GetLastError function.

From that I can only conclude they are the same. Should I changed it to `io::Error::last_error` with a comment saying it's equal to `WSAGetLastError`?

stjepang: It's fine either way.

I was just concerned that accidentally using `syscall!()` on a non-WSA function may introduce bugs. But it won't, so we're good.

Thomasdezeeuw: Changed it to use `io::Error::last_os_error` with the references above in the commit message.

Issue #113, opened 2020-10-09
Expose all flags in msghdr send/recv_vectored function


Thomasdezeeuw:
Pr #107 added vectored I/O and currently returns a boolean indicating if the buffer was truncated. But the `flags` field on `msghdr` give us more information. This issue is about nicely (and cross-platform) exposing that information.

Some ideas:
* Add our own `msghdr` like structure that can be checked after a call to send/recv. Downside: input, output arguments are very "rustic".
* Instead of returning a boolean indicating truncation, change it to the entire `flags` field wrapped in a new type. Such a new type could be `SendFlags` with function such as `is_truncated(&self) -> bool`.

/cc @de-vri-es

de-vri-es:
I like the flags type idea. It also makes the return type more intuitive: `(usize, RecvFlags, SockAddr)` instead of `(usize, bool SockAdd)`.

There are no flags to return when sending messages, btw. But I think `RecvFlags` is still a good name. It would allow to create a `SendFlags` type later for the flags argument.

It probably makes sense to do this before releasing vectored I/O, to prevent backwards incompatibilities. Initially it could even just expose `is_truncated()` until support for more flags is added.

Implementing this on Windows will be a bit harder. The flags are only used to report 'MSG_PARTIAL`, which is not the same as `MSG_TRUNC`. Truncated messages are reported as error, which can be turned into a flag. But the `MSG_EOR` and `MSG_OOB` flags aren't reported at all.

For `MSG_EOR` that doesn't matter: Windows has no connection oriented datagram sockets, so the flag wouldn't be reported anyway. I don't see a way to detect `MSG_OOB` on Windows. It appears that it can only be used as input flag.

de-vri-es:
PR opened at #114 :)

de-vri-es:
#114 was merged :)

Do you think that there is more that should be done?

Thomasdezeeuw:
Nope.

PR #114, opened 2020-10-10
Return new `RecvFlags` struct with vectored I/O.

de-vri-es:
The `RecvFlags` struct replaces the boolean to indicate if messages are truncated. This is intended to address #113.

Currently it only supports POSIX flags on Unix, no Linux specific flags, although that would be trivial to add.

On Windows, it only supports `is_truncated()`, because the other flags don't exist.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -247,29 +247,6 @@ pub struct RecvFlags(c_int);
 
 #[cfg(not(target_os = "redox"))]
 impl RecvFlags {
-    /// Check if the message terminates a record.
-    ///
-    /// Not all socket types support the notion of records.
-    /// For socket types that do support it (such as [`SEQPACKET`][Type::SEQPACKET]),
-    /// a record is terminated by sending a message with the end-of-record flag set.
-    ///
-    /// On Unix this corresponds to the MSG_EOR flag.
-    #[cfg(unix)]
-    pub fn is_end_of_record(self) -> bool {
-        self.0 & sys::MSG_EOR != 0
-    }
-
-    /// Check if the message contains out-of-band data.
-    ///
-    /// This is useful for protocols where you receive out-of-band data
-    /// mixed in with the normal data stream.
-    ///
-    /// On Unix this corresponds to the MSG_OOB flag.
-    #[cfg(unix)]
-    pub fn is_out_of_band(self) -> bool {
-        self.0 & sys::MSG_OOB != 0
-    }
-
     /// Check if the message contains a truncated datagram.
     ///
     /// This flag is only used for datagram-based sockets,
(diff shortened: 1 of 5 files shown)


de-vri-es:
Sorry about the force pushing. I think nobody looked yet. Further changes will be separate commits that can be squashed later.

src/sys/windows.rs
```diff
@@ -42,6 +42,9 @@ const WSA_FLAG_OVERLAPPED: DWORD = 0x01;
 
 pub use winapi::ctypes::c_int;
 
+/// Fake MSG_TRUNC flag for the [`RecvFlags`] struct.
+pub(crate) const MSG_TRUNC: c_int = sock::WSAEMSGSIZE;
```
de-vri-es, line 46: Wasn't too sure about this: using the error code as "flag". Probably a bad idea now that I think about it. It's highly unlikely that it is a power of two. I'll just copy the value from Linux instead.

de-vri-es: Replaced with 0x20 and a comment that it's copied from Linux.

src/lib.rs
```diff
@@ -237,3 +237,60 @@ impl From<Protocol> for c_int {
...
+#[cfg(not(target_os = "redox"))]
+impl RecvFlags {
+    /// Check if the message terminates a record.
+    ///
+    /// Not all socket types support the notion of records.
+    /// For socket types that do support it (such as [`SEQPACKET`][Type::SEQPACKET]),
+    /// a record is terminated by sending a message with the end-of-record flag set.
+    ///
+    /// On Unix this corresponds to the MSG_EOR flag.
+    #[cfg(unix)]
```
Thomasdezeeuw, line 257: Can you move this and `is_out_of_band` to the sys/unix.rs file, similar to e.g. `Domain` extension: https://github.com/alexcrichton/socket2-rs/blob/d653b7b5e0b2438677b721508f4ba5322a9bd81a/src/sys/unix.rs#L98-L110

src/sys/windows.rs
```diff
@@ -42,6 +42,11 @@ const WSA_FLAG_OVERLAPPED: DWORD = 0x01;
 
 pub use winapi::ctypes::c_int;
 
+/// Fake MSG_TRUNC flag for the [`RecvFlags`] struct.
```
Thomasdezeeuw, line 45: Could you add a comment that this is used if `WSAEMSGSIZE` is returned?

src/sys/windows.rs
```diff
@@ -42,6 +42,11 @@ const WSA_FLAG_OVERLAPPED: DWORD = 0x01;
...
+///
+/// Value copied from Linux.
```
Thomasdezeeuw, line 47: I think you can just say the value is defined by us and just use `0b1`.

Thomasdezeeuw (commented):
LGTM expect for some small things.

src/lib.rs
de-vri-es: Moved, along with the `Debug` impl.

src/sys/windows.rs
de-vri-es: Added :)

src/sys/windows.rs
de-vri-es: Done :)

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @de-vri-es.

Issue #111, opened 2020-10-08
Bikeshed: name for Socket::new_with_common_flags


Thomasdezeeuw:
Pr #110 changed `Socket::new` to be a simple call to `socket(2)`, with setting common flags like `CLO_EXEC`. For convenience we should add back a function which copies the old behaviour. For that function we need a good name. Some possibilities:

* `Socket::new_with_common_flags`: too long.
* `Socket::like_std`: create a socket like the standard library would?

Thomasdezeeuw:
We should use `cfg(accessible)` for this once stable: https://github.com/rust-lang/rust/issues/64797.

de-vri-es:
Hmm, I understand the reasoning behind this, but I do think that 9 out of 10 times (if not more) you want to use the default flags.

So I would actually flip this: have `Socket::new()` use the almost always desired flags, and have `Socket::with_custom_flags` for the other use-cases.

Thomasdezeeuw:
> Hmm, I understand the reasoning behind this, but I do think that 9 out of 10 times (if not more) you want to use the default flags.

This is probably true. 

> So I would actually flip this: have `Socket::new()` use the almost always desired flags, and have `Socket::with_custom_flags()` for the other use-cases.

I still argue for keeping `Socket::new` as a wrapper for `socket(2)`. I see the crate socket2  as a small wrapper around the libc interface with some convenient functions added, such as `Domain::for_address`, and I would argue that `Socket::new_with_common_flags` is one of those convenience functions.

de-vri-es:
Hmm, another reason to be careful is that it will change the behavior of existing code. Currently, sockets are always created with the `CLOEXEC` flag set. After the change the file descriptors will be leaked to child processes if the default `new()` is used. Of course there is semver to signify breaking changes, but most people will just fix compilation errors rather than look in=depth what changed.

To me personally, it also feels wrong to create file descriptors without close-on-exec set by default. Leaking file descriptors is bad, so not a nice default.

Also, currently there isn't even a way to set the close-on-exec flag after creation for the user. At the very least that should be exposed for the user, so `with_default_flag()` doesn't do magic the user can't reproduce without fiddling with the raw fd.

Thomasdezeeuw:
[...]

PR #117, opened 2020-10-30
Set close-on-exec and similar by default.

de-vri-es:
This is my proposal for #111: the default behavior is to enable close-on-exec and `SOCK_NOSIGPIPE` on the relevant platforms, but there are `*_raw` functions that really only wrap a single syscall.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -50,41 +50,6 @@ use crate::{Domain, Protocol, SockAddr, Type};
 /// can lead to a data race when two threads are changing options in parallel.
 ///
 /// # Examples
-///
-/// Creating a new socket setting all advisable flags.
-///
-#[cfg_attr(feature = "all", doc = "```")] // Protocol::cloexec requires the `all` feature.
-#[cfg_attr(not(feature = "all"), doc = "```ignore")]
-/// # fn main() -> std::io::Result<()> {
-/// use socket2::{Protocol, Domain, Type, Socket};
-///
-/// let domain = Domain::IPV4;
-/// let ty = Type::STREAM;
-/// let protocol = Protocol::TCP;
-///
-/// // On platforms that support it set `SOCK_CLOEXEC`.
-/// #[cfg(any(target_os = "android", target_os = "dragonfly", target_os = "freebsd", target_os = "linux", target_os = "netbsd", target_os = "openbsd"))]
-/// let ty = ty.cloexec();
-///
-/// // On windows set `WSA_FLAG_NO_HANDLE_INHERIT`.
-/// #[cfg(windows)]
-/// let ty = ty.no_inherit();
-///
-/// let socket = Socket::new(domain, ty, Some(protocol))?;
-///
-/// // On platforms that don't support `SOCK_CLOEXEC`, use `FD_CLOEXEC`.
-/// #[cfg(all(not(windows), not(any(target_os = "android", target_os = "dragonfly", target_os = "freebsd", target_os = "linux", target_os = "netbsd", target_os = "openbsd"))))]
-/// socket.set_cloexec(true)?;
-///
-/// // On macOS and iOS set `NOSIGPIPE`.
-/// #[cfg(target_vendor = "apple")]
-/// socket.set_nosigpipe(true)?;
-///
-/// # drop(socket);
-/// # Ok(())
-/// # }
-/// ```
-///
 /// ```no_run
 /// # fn main() -> std::io::Result<()> {
 /// use std::net::{SocketAddr, TcpListener};
@@ -112,27 +77,51 @@ pub struct Socket {
 impl Socket {
     /// Creates a new socket ready to be configured.
     ///
+    /// This function corresponds to `socket(2)` on Unix and `WSASocketW` on Windows.
+    ///
+    /// # Notes
+    ///
+    /// On Unix-like systems, the close-on-exec flag is set on the new socket.
+    /// Additionally, on macOS X and iOS the `SOCK_NOSIGPIPE` option is set.
+    /// On Windows, the socket handle is made non-inheritable.
+    ///
+    /// If possible, the close-on-exec flag will be set atomically when the socket is created.
+    /// However, on some platforms the flag is to be set with a separate syscall,
+    /// which leaves a small window for race conditions.
+    ///
+    /// Use [`Socket::new_raw`] if you don't want these flags and options to be set.
+    pub fn new(domain: Domain, ty: Type, protocol: Option<Protocol>) -> io::Result<Socket> {
+        // On platforms that support it set `SOCK_CLOEXEC` atomically.
+        #[cfg(any(
+            target_os = "android",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "linux",
+            target_os = "netbsd",
+            target_os = "openbsd",
+        ))]
+        let ty = ty._cloexec();
+
+        // On windows set `NO_HANDLE_INHERIT` atomically.
+        #[cfg(windows)]
+        let ty = ty._no_inherit();
+
+        set_common_flags(Socket::new_raw(domain, ty, protocol)?)
+    }
+
+    /// Creates a new socket ready to be configured.
+    ///
     /// This function corresponds to `socket(2)` on Unix and `WSASocketW` on
     /// Windows and simply creates a new socket, no other configuration is done
     /// and further functions must be invoked to configure this socket.
     ///
     /// # Notes
     ///
-    /// The standard library sets the `CLOEXEC` flag on Unix on sockets, this
-    /// function does **not** do this, but its advisable. On supported platforms
-    /// [`Type::cloexec`] can be used for this, or by using
-    /// [`Socket::set_cloexec`].
-    ///
-    /// Furthermore on macOS and iOS `NOSIGPIPE` is not set, this can be done
-    /// using [`Socket::set_nosigpipe`].
-    ///
-    /// Similarly on Windows the `HANDLE_FLAG_INHERIT` is **not** set to zero,
-    /// but again in most cases its advisable to do so. This can be doing using
-    /// [`Socket::set_no_inherit`].
-    ///
-    /// See the `Socket` documentation for a full example of setting all the
-    /// above mentioned flags.
-    pub fn new(domain: Domain, ty: Type, protocol: Option<Protocol>) -> io::Result<Socket> {
+    /// This function does not set the close-on-exec flag or the `SOCK_SIGNOPIPE` option
+    /// for the created socket. On Windows, the created socket handle will be inheritable.
+    /// You should normally use [`Socket::new`], which creates sockets with the advised
+    /// flags and options for the current platform.
+    pub fn new_raw(domain: Domain, ty: Type, protocol: Option<Protocol>) -> io::Result<Socket> {
         let protocol = protocol.map(|p| p.0).unwrap_or(0);
         sys::socket(domain.0, ty.0, protocol).map(|inner| Socket { inner })
     }
@@ -143,12 +132,52 @@ impl Socket {
     ///
     /// # Notes
     ///
-    /// Much like [`Socket::new`] this doesn't set any flags, which might be
-    /// advisable.
+    /// On Unix-like systems, the close-on-exec flag is set on the new sockets.
+    /// Additionally, on macOS X and iOS the `SOCK_NOSIGPIPE` option is set.
+    /// On Windows, the socket handles are made non-inheritable.
+    ///
+    /// If possible, the close-on-exec flag will be set atomically when the socket is created.
+    /// However, on some platforms the flag is to be set with a separate syscall,
+    /// which leaves a small window for race conditions.
+    ///
+    /// Use [`Socket::pair_raw`] if you don't want these flags and options to be set.
     ///
     /// This function is only available on Unix.
     #[cfg(all(feature = "all", unix))]
     pub fn pair(
+        domain: Domain,
+        ty: Type,
+        protocol: Option<Protocol>,
+    ) -> io::Result<(Socket, Socket)> {
+        // On platforms that support it set `SOCK_CLOEXEC` atomically.
+        #[cfg(any(
+            target_os = "android",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "linux",
+            target_os = "netbsd",
+            target_os = "openbsd",
+        ))]
+        let ty = ty._cloexec();
+
+        let (a, b) = Socket::pair_raw(domain, ty, protocol)?;
+        Ok((set_common_flags(a)?, set_common_flags(b)?))
+    }
+
+    /// Creates a pair of sockets which are connected to each other.
+    ///
+    /// This function corresponds to `socketpair(2)`.
+    ///
+    /// # Notes
+    ///
+    /// This function does not set the close-on-exec flag or the `SOCK_SIGNOPIPE` option
+    /// for the created sockets. You should normally use [`Socket::pair`],
+    /// which creates sockets with the advised
+    /// flags and options for the current platform.
+    ///
+    /// This function is only available on Unix.
+    #[cfg(all(feature = "all", unix))]
+    pub fn pair_raw(
         domain: Domain,
         ty: Type,
         protocol: Option<Protocol>,
@@ -198,6 +227,50 @@ impl Socket {
 
     /// Accept a new incoming connection from this listener.
     ///
+    /// This function corresponds to the `accept(2)` function on
+    /// Windows and Unix.
+    ///
+    /// # Notes
+    ///
+    /// On Unix-like systems, the close-on-exec flag is set on the new socket.
+    /// Additionally, on macOS X and iOS the `SOCK_NOSIGPIPE` option is set.
+    /// On Windows, the socket handle is made non-inheritable.
+    ///
+    /// If possible, the close-on-exec flag will be set atomically when the socket is created.
+    /// However, on some platforms the flag is to be set with a separate syscall,
+    /// which leaves a small window for race conditions.
+    ///
+    /// Use [`Socket::accept_raw`] if you don't want these flags and options to be set.
+    pub fn accept(&self) -> io::Result<(Socket, SockAddr)> {
+        // On platforms that support it set `SOCK_CLOEXEC` atomically.
+        #[cfg(any(
+            target_os = "android",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "linux",
+            target_os = "netbsd",
+            target_os = "openbsd",
+        ))]
+        let (sock, addr) = self._accept4(libc::SOCK_CLOEXEC)?;
+
+        #[cfg(not(any(
+            target_os = "android",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "linux",
+            target_os = "netbsd",
+            target_os = "openbsd",
+        )))]
+        let (sock, addr) = self.accept_raw()?;
+
+        #[cfg(windows)]
+        sock._set_no_inherit(true)?;
+
+        Ok((sock, addr))
+    }
+
+    /// Accept a new incoming connection from this listener.
+    ///
     /// This function directly corresponds to the `accept(2)` function on
     /// Windows and Unix.
     ///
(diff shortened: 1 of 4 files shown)


de-vri-es:
Note that this PR doesn't re-add the workaround for missing SOCK_CLOEXEC support, in line with https://github.com/rust-lang/rust/pull/74606. That keeps the code much simpler.

Thomasdezeeuw:
I'm not sure about the `raw`, I want to think about this a bit more, but currently don't have the time for it.

de-vri-es:
Some alternatives I can think of: `plain`, `custom`, `with_custom_flags`, `uncooked`, `transparent`.

Although out of all of those, `raw` still sounds best to me.

Thomasdezeeuw:
I don't want to accept this as is. With `Type::no_inherit` from pr #124, this can be completely handled in `src/socket.rs`, essentially copying the example code (removed in this pr) into a method. I want the `sys` module be only methods that are direct wrappers around system calls with minimal overhead and some helper functions.

I'll change `Socket::new` to the method I decided above.

de-vri-es:
> I don't want to accept this as is. With `Type::no_inherit` from pr #124, this can be completely handled in `src/socket.rs`, essentially copying the example code (removed in this pr) into a method. I want the `sys` module be only methods that are direct wrappers around system calls with minimal overhead and some helper functions.
> 
> I'll change `Socket::new` to the method I decided above.

That's not true though, at-least not without putting more `#[cfg(...)]` in `src/socket.rs`, since you would have to enable platform specific flags in the calls. It would also requires the `all` feature just for correct initialization. That kinda defeats the purpose of having `sys/unix.rs` and `sys/windows.rs`.

It's also not a fix for `accept4`, which again is platform specific.

Thomasdezeeuw:
@de-vri-es I created pr #125 for `Socket::new`.

Thomasdezeeuw:
The relevant commit is https://github.com/rust-lang/socket2-rs/pull/125/commits/fdc4fd47a93435ab27bf20b8d4e9e47eb9ad50c8.

de-vri-es:
I'm having a hard time contributing this way. I can understand feedback and requests for changes, but now you're leaving my PR open without feedback for 11 days and then break my PR without letting me implement feedback.

I would appreciate a more contribution friendly workflow :(

Thomasdezeeuw:
> I'm having a hard time contributing this way. I can understand feedback and requests for changes, but now you're leaving my PR open without feedback for 11 days and then break my PR without letting me implement feedback.

I'm sorry about this. As I said I wanted to this think about the pr (and the bigger picture of flags set to et etc.) a bit more, but I've been rather busy and socket2 is simply not among my highest priorities. This is even further complicated because I'm in the process of refactoring the entire crate, of which only roughly 20% is currently done.

> I would appreciate a more contribution friendly workflow :(

I'll try to improve this in the future, hopefully once the rewrite is done this will be much improved. I did something similar with Mio, but I did the refactoring offline and only contributed the changes back once it was complete. Perhaps it was a mistake to take a different approach to socket2.

de-vri-es:
I'm sorry to hear that. If the refactor is not finished though, why can't you review this and give me time to implement feedback before continuing with the refactor? There is no need to rush anything, including the review. It's just the new PRs that you're opening that make it impossible for me to contribute.

Especially if you're pressed for time, you can let me do the legwork..

Thomasdezeeuw:
> I'm sorry to hear that. If the refactor is not finished though, why can't you review this and give me time to implement feedback before continuing with the refactor? There is no need to rush anything, including the review. It's just the new PRs that you're opening that make it impossible for me to contribute.

I created pr #125 just as an example, it's not complete, it doesn't even compile. I just wanted to show the direction I was thinking in, it's often easier to show with (pseude/broken) code that in English for me.

> Especially if you're pressed for time, you can let me do the legwork..

If you can help review #124 and rebase this pr on it with, with the suggested changes in line with #125 that would be great.

de-vri-es:
Ah, that does sound much better. I'll make some time for that :)

de-vri-es:
Ok, rebased and process the suggestions.

The PR also hides the public `set_cloexec()` and `set_no_inherit()` behind the `all` feature, for consistency with other platform specific APIs.

Platform specific APIs needed by `new()` and `accept()` now have a `pub(crate)` variant that starts with an underscore. `pair()` is already gated behind the `all` feature, so it wasn't needed for that.

src/socket.rs
```diff
@@ -110,6 +75,54 @@ pub struct Socket {
...
+            target_os = "openbsd",
+        ))]
+        {
+            Self::new_raw(domain, ty._cloexec(), protocol)
+        }
+
+        #[cfg(windows)]
+        {
+            Self::new_raw(domain, ty._no_inherit(), protocol)
+        }
```
Thomasdezeeuw, line 109: Let's keep a single call to `Socket::new_raw` (please don't use `Self`). Much like the example just use `let ty = ty._cloexec()`.

src/socket.rs
```diff
@@ -110,6 +75,54 @@ pub struct Socket {
...
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "linux",
+            target_os = "netbsd",
+            target_os = "openbsd",
+            target_os = "windows",
+        )))]
+        {
+            // Fix portability flags after creating the socket.
+            fixup_socket(Self::new_raw(domain, ty, protocol)?)
```
Thomasdezeeuw, line 122: `fixup` is not a good name as it indicates the socket is somehow broken.

src/socket.rs
```diff
@@ -118,23 +131,62 @@ impl Socket {
     ///
     /// # Notes
     ///
-    /// The standard library sets the `CLOEXEC` flag on Unix on sockets, this
-    /// function does **not** do this, but its advisable. On supported platforms
-    /// [`Type::cloexec`] can be used for this, or by using
-    /// [`Socket::set_cloexec`].
+    /// This function does not set the close-on-exec flag or the `SOCK_SIGNOPIPE` option
+    /// for the created socket. On Windows, the created socket handle will be inheritable.
+    /// You should normally use [`Self::new`], which creates sockets with the advised
```
Thomasdezeeuw, line 136: Not `Self::new`, but `Socket::new`.

src/socket.rs
```diff
@@ -143,19 +195,21 @@ impl Socket {
...
+    pub fn pair_raw(
         domain: Domain,
         ty: Type,
         protocol: Option<Protocol>,
     ) -> io::Result<(Socket, Socket)> {
         let protocol = protocol.map(|p| p.0).unwrap_or(0);
-        sys::socketpair(domain.0, ty.0, protocol)
-            .map(|fds| (Socket { inner: fds[0] }, Socket { inner: fds[1] }))
+        let [a, b] = sys::socketpair(domain.0, ty.0, protocol)?;
+        Ok((Self { inner: a }, Self { inner: b }))
```
Thomasdezeeuw, line 212: Please revert this, not needed.

src/socket.rs
```diff
@@ -143,19 +195,21 @@ impl Socket {
     ///
     /// # Notes
     ///
-    /// Much like [`Socket::new`] this doesn't set any flags, which might be
-    /// advisable.
+    /// This function does not set the close-on-exec flag or the `SOCK_SIGNOPIPE` option
+    /// for the created sockets. You should normally use [`Self::pair`],
```
Thomasdezeeuw, line 199: Not `Self::pair`, `Socket::pair`.

src/socket.rs
```diff
@@ -1017,6 +1119,28 @@ impl Drop for Socket {
...
+        // Even though new sockets will have the no-inherit flag set,
+        // sockets obtained through [`Socket::accept_raw()`] are
+        // still inheritable on Windows.
+        sock._set_no_inherit(true)?;
+    }
+    #[cfg(unix)]
+    {
+        sock._set_cloexec(true)?;
+    }
+    #[cfg(all(target_vendor = "apple"))]
```
Thomasdezeeuw, line 1136: Not need for `all`.

src/socket.rs
```diff
@@ -1017,6 +1119,28 @@ impl Drop for Socket {
     }
 }
 
+/// Fix-up a created socket by setting the proper portability flags.
+#[allow(unused)]
+fn fixup_socket(mut sock: Socket) -> io::Result<Socket> {
```
Thomasdezeeuw, line 1124: Better name would be `set_common_flags` or something, but `fixup` shouldn't be in it.

src/socket.rs
```diff
@@ -1017,6 +1119,28 @@ impl Drop for Socket {
...
+    #[cfg(windows)]
+    {
+        // Even though new sockets will have the no-inherit flag set,
+        // sockets obtained through [`Socket::accept_raw()`] are
+        // still inheritable on Windows.
+        sock._set_no_inherit(true)?;
+    }
+    #[cfg(unix)]
```
Thomasdezeeuw, line 1132: Too broad, needs to be the inverse of `Type::set_closexec`. Don't want to make unneeded system calls.

src/socket.rs
```diff
@@ -1017,6 +1119,28 @@ impl Drop for Socket {
     }
 }
 
+/// Fix-up a created socket by setting the proper portability flags.
+#[allow(unused)]
+fn fixup_socket(mut sock: Socket) -> io::Result<Socket> {
+    #[cfg(windows)]
+    {
```
Thomasdezeeuw, line 1126: Move this to `accept`, not needed in `Socket::new` or `Socket::new_pair`.

src/socket.rs
```diff
@@ -1017,6 +1119,28 @@ impl Drop for Socket {
     }
 }
 
+/// Fix-up a created socket by setting the proper portability flags.
```
Thomasdezeeuw, line 1122: These flags aren't for portability. These flags are avoiding a possible resource leak (inherit/clos-exec) and unhandled signals (sigpipe).

src/socket.rs
```diff
@@ -207,11 +261,59 @@ impl Socket {
...
+    /// remote peer's address will be returned.
+    ///
+    /// # Notes
+    ///
+    /// This function does not set the close-on-exec flag or the `SOCK_SIGNOPIPE` option
+    /// for the created socket. On Windows, the created socket handle will be inheritable.
+    /// You should normally use [`Self::accept_raw`], which creates sockets with the advisable
+    /// flags and options for the current platform.
+    pub fn accept_raw(&self) -> io::Result<(Socket, SockAddr)> {
+        sys::accept(self.inner).map(|(inner, addr)| (Self { inner }, addr))
```
Thomasdezeeuw, line 316: Revert this line, not needed.

src/sys/unix.rs
```diff
@@ -331,6 +343,31 @@ pub(crate) fn accept(fd: SysSocket) -> io::Result<(SysSocket, SockAddr)> {
...
+#[cfg(any(
+    target_os = "android",
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "illumos",
+    target_os = "linux",
+    target_os = "netbsd",
+    target_os = "openbsd",
+))]
+pub(crate) fn accept4(fd: SysSocket, flags: c_int) -> io::Result<(SysSocket, SockAddr)> {
```
Thomasdezeeuw, line 355: Move this into `_accept4`, same as with `Socket::set_cloexec`.

src/sys/unix.rs
```diff
@@ -436,12 +475,17 @@ impl crate::Socket {
...
+    #[cfg(all(target_vendor = "apple"))]
+    pub(crate) fn _set_nosigpipe(&self, nosigpipe: bool) -> io::Result<()> {
         unsafe {
-            setsockopt::<c_int>(
+            setsockopt(
                 self.inner,
                 libc::SOL_SOCKET,
                 libc::SO_NOSIGPIPE,
-                nosigpipe as _,
+                nosigpipe as c_int,
```
Thomasdezeeuw, line 488: Let's leave this as `_`. Apple has a history of change system calls parameters, with the old code was forward compatible in that way.

src/sys/unix.rs
```diff
@@ -436,12 +475,17 @@ impl crate::Socket {
     /// Only supported on Apple platforms (`target_vendor = "apple"`).
     #[cfg(all(feature = "all", target_vendor = "apple"))]
     pub fn set_nosigpipe(&self, nosigpipe: bool) -> io::Result<()> {
+        self._set_nosigpipe(nosigpipe)
+    }
+
+    #[cfg(all(target_vendor = "apple"))]
```
Thomasdezeeuw, line 481: No  `all` required.

src/sys/unix.rs
```diff
@@ -503,15 +547,6 @@ unsafe fn setsockopt<T>(fd: SysSocket, opt: c_int, val: c_int, payload: T) -> io
     .map(|_| ())
 }
 
-/*
```
Thomasdezeeuw, line 506: Whoops left this in...

src/sys/unix.rs
```diff
@@ -490,7 +534,7 @@ unsafe fn getsockopt<T>(fd: SysSocket, opt: c_int, val: c_int) -> io::Result<T>
 }
 
 /// Caller must ensure `T` is the correct type for `opt` and `val`.
-#[cfg(all(feature = "all", target_vendor = "apple"))]
+#[cfg(target_vendor = "apple")]
```
Thomasdezeeuw, line 537: Are you sure this won't cause warnings? I'm quite sure I added it for a reason.

src/sys/unix.rs
```diff
@@ -1018,6 +1053,12 @@ impl Socket {
     }
 }
 
+impl Drop for Socket {
```
Thomasdezeeuw, line 1056: Revert this; now `crate::Socket` is closed twice.

src/sys/windows.rs
```diff
@@ -882,6 +891,12 @@ impl Socket {
     }
 }
 
+impl Drop for Socket {
```
Thomasdezeeuw, line 894: Revert this; now `crate::Socket` is closed twice.

src/socket.rs
```diff
@@ -207,11 +261,59 @@ impl Socket {
...
+            Ok((fixup_socket(sock)?, addr))
+        }
+    }
+
+    /// Accept a new incoming connection from this listener.
+    ///
+    /// This function directly corresponds to the `accept(2)` function on
+    /// Windows and Unix.
+    ///
+    /// This function will block the calling thread until a new connection is
```
Thomasdezeeuw, line 305: Remove this section, this is not true if the socket is set to non-blocking mode.

src/socket.rs
```diff
@@ -118,23 +131,62 @@ impl Socket {
...
+        #[cfg(any(
+            target_os = "android",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "linux",
+            target_os = "netbsd",
+            target_os = "openbsd",
+        ))]
+        {
+            Self::pair_raw(domain, ty._cloexec(), protocol)
```
Thomasdezeeuw, line 175: No `Self`.

src/socket.rs
```diff
@@ -118,23 +131,62 @@ impl Socket {
...
+        #[cfg(not(any(
+            target_os = "android",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "linux",
+            target_os = "netbsd",
+            target_os = "openbsd",
+        )))]
+        {
+            let (a, b) = Self::pair_raw(domain, ty, protocol)?;
```
Thomasdezeeuw, line 187: No `Self`.

src/socket.rs
```diff
@@ -118,23 +131,62 @@ impl Socket {
...
-    /// function does **not** do this, but its advisable. On supported platforms
-    /// [`Type::cloexec`] can be used for this, or by using
-    /// [`Socket::set_cloexec`].
+    /// This function does not set the close-on-exec flag or the `SOCK_SIGNOPIPE` option
+    /// for the created socket. On Windows, the created socket handle will be inheritable.
+    /// You should normally use [`Self::new`], which creates sockets with the advised
+    /// flags and options for the current platform.
+    pub fn new_raw(domain: Domain, ty: Type, protocol: Option<Protocol>) -> io::Result<Socket> {
+        let protocol = protocol.map(|p| p.0).unwrap_or(0);
+        sys::socket(domain.0, ty.0, protocol).map(|inner| Self { inner })
```
Thomasdezeeuw, line 140: Revert this line.

src/socket.rs
```diff
@@ -110,6 +75,54 @@ pub struct Socket {
...
+    /// On Unix-like systems, the close-on-exec flag is set on the new socket.
+    /// Additionally, on macOS X and iOS the `SOCK_NOSIGPIPE` option is set.
+    /// On Windows, the socket handle is made non-inheritable.
+    ///
+    /// If possible, the close-on-exec flag will be set atomically when the socket is created.
+    /// However, on some platforms the flag is to be set with a separate syscall,
+    /// which leaves a small window for race conditions.
+    ///
+    /// Use [`Self::new_raw`] if you don't want these flags and options to be set.
+    pub fn new(domain: Domain, ty: Type, protocol: Option<Protocol>) -> io::Result<Socket> {
```
Thomasdezeeuw, line 93: Please use the same structure I used here: https://github.com/rust-lang/socket2-rs/pull/125/files#diff-917ff224180b881299cdcef7bc14e710509229092ea4e6599861680bc0a2a458R82-R117, comments included.

Thomasdezeeuw (changes requested):
I'm afraid this isn't quite what I was looking for. I'm added a bunch of inline comments.

Also please don't use the `Self` key when returning or creating type, it requires someone to needlessly look up what type is being created, just use the concrete type name.

src/socket.rs
de-vri-es: Well, I'd consider not having to worry about these platform specific things makes it portable, but fine.

src/socket.rs
de-vri-es: Which syscall here is unneeded?

src/socket.rs
de-vri-es: It is not called in `Socket::new()` on windows, and `new_pair()` doesn't exist on Windows. Doing it here keeps all the platform specific fix-ups in one place. But sure, I'll move it.

src/socket.rs
de-vri-es: This is an artifact of having to rebase my PR on your PR, but sure, I will.

src/sys/unix.rs
de-vri-es: It's called in `_set_nosighup` now. CI would catch the warnings.

src/socket.rs
de-vri-es: I'd argue it is without those flags, but sure, will rename.

src/socket.rs
de-vri-es: Sure, that works too.

src/socket.rs
de-vri-es: Ah, I see. This was never called on platforms that don't need to set cloexec separately. But sure, we can always call it and move the check in here.

src/sys/unix.rs
de-vri-es: It's not though. `crate::Socket` doesn't contain a `sys::Socket`. And `sys::Socket` implements `IntoRawFd`, so it should own it's file descriptor. Or it should not implement `IntoRawFd`.

From the docs of `IntoRawFd`:
> A trait to express the ability to consume an object and acquire ownership of its raw file descriptor.

If `sys::Socket` doesn't own it's filedescriptor, it can't make that promise.

src/sys/windows.rs
de-vri-es: Same as above.

src/sys/unix.rs
de-vri-es: It's not a functional change, I just moved the `c_int` from the template parameter to the cast, in line with the rest of the code in this file.

We must be specific with the type here.

src/socket.rs
de-vri-es: renamed

de-vri-es:
> I'm afraid this isn't quite what I was looking for.

You could just leave this bit off. The in-line comments are quite clear on their own, and that kind of remark makes me feel somewhat unappreciated.

Anyway, I think I processed most of the comments. I marked the processed ones as resolved, and we probably need to discuss the remaining ones a bit more.

Thomasdezeeuw:
@de-vri-es I've rebased this on my branch in #134.

de-vri-es:
I'm sorry, but I'm done here. I can't contribute to a repository where the maintainer does things like this. 
Even though I explicitly brought it up before, you still went ahead and broke my PR for a third time, leaving this without review again for 11 days.

You may have rebased my PR, but you also modified it. Now it's up to me to search for modifications and comment if I don't agree with the changes.

Finally, I would hugely appreciate it if you use a more human friendly tone in your feedback. I'd much prefer "Could you X", "I'd prefer Y" over "Revert X", "Do Y". I'm not a robot after all.

I wish you the best of luck with this repo, but I can't contribute to it like this.

Thomasdezeeuw:
> I'm sorry, but I'm done here. I can't contribute to a repository where the maintainer does things like this.
> Even though I explicitly brought it up before, you still went ahead and broke my PR for a third time, leaving this without review again for 11 days.

Really... we've been through this. I don't have all the time in the world you review your pr.

> You may have rebased my PR, but you also modified it. Now it's up to me to search for modifications and comment if I don't agree with the changes.

It's very common to accept a PR with changes. I did it in the same commit, but I also could have made it another commit, really who cares?

> Finally, I would hugely appreciate it if you use a more human friendly tone in your feedback. I'd much prefer "Could you X", "I'd prefer Y" over "Revert X", "Do Y". I'm not a robot after all.

I've been doing too long to say please, especially since a thank you is rarely around the corner. I think you're taking these things way to personal. I've always kept my comments about your code and never made it personal, I can't say the same about your comments (which made it tiring you work you to be honest, but I never mentioned that did I?).

> I wish you the best of luck with this repo, but I can't contribute to it like this.

Thank you. Even though we might not see eye to eye now I also wish you the best of luck.

PR #118, opened 2020-11-05
add test for roundtripping TCP keepalive duration

hawkw:
This PR adds a test for round-tripping TCP keepalive time.

I believe the current implementation of `Socket::keepalive` on Windows
is incorrect, so I expect this to fail on Windows CI. This PR is 
intended primarily to confirm that the current implementation doesn't
work (although it could be merged once the issue is fixed).

Signed-off-by: Eliza Weisman <eliza@buoyant.io>

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -301,3 +301,13 @@ fn recv_from_vectored_truncated() {
     assert_eq!(addr.as_inet6().unwrap(), addr_a.as_inet6().unwrap());
     assert_eq!(&buffer, b"do not feed the gremlins");
 }
+
+#[test]
+fn keepalive_roundtrips() {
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
@hawkw I missed this pr, sorry. But you're right Windows CI failed.

Thomasdezeeuw:
This is now implemented here: https://github.com/rust-lang/socket2/blob/819716c66f6c591388fb2ec117503f96f05f15d2/tests/socket.rs#L549-L612 and https://github.com/rust-lang/socket2/blob/819716c66f6c591388fb2ec117503f96f05f15d2/tests/socket.rs#L734.

PR #120, opened 2020-11-06
 Don't assume memory layout of std::net::SocketAddr

### LINKED ISSUE #119: The code assumes the memory layout of SocketAddrV{4,6}
This library casts `std::net::SocketAddrV4` (and `V6`) into `libc::sockaddr`: https://github.com/rust-lang/socket2-rs/blob/b0f77842b3357dd571bb86b30b354bc55215f693/src/sockaddr.rs#L95-L115

As far as I can tell there are no guarantees from `std` about the layout of `SocketAddrV{4,6}`, and this code could silently compile and cause UB elsewhere if the representation changes.

This internals forum thread is where this discussion started: https://internals.rust-lang.org/t/why-are-socketaddrv4-socketaddrv6-based-on-low-level-sockaddr-in-6/13321

`mio` does the same kind of invalid casting: https://github.com/tokio-rs/mio/issues/1386

faern:
Fixes #119

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ features = ["handleapi", "ws2def", "ws2ipdef", "ws2tcpip", "minwindef"]
 
 [target."cfg(any(unix, target_os = \"redox\"))".dependencies]
 cfg-if = "0.1.6"
-libc = "0.2.66"
+libc = { version = "0.2.66", features = ["align"] }
 
 [target."cfg(target_os = \"redox\")".dependencies]
 redox_syscall = "0.1.38"
--- a/src/sockaddr.rs
+++ b/src/sockaddr.rs
@@ -5,9 +5,13 @@ use std::ptr;
 
 #[cfg(any(unix, target_os = "redox"))]
 use libc::{
-    sa_family_t, sockaddr, sockaddr_in, sockaddr_in6, sockaddr_storage, socklen_t, AF_INET,
-    AF_INET6,
+    in6_addr, in_addr, sa_family_t, sockaddr, sockaddr_in, sockaddr_in6, sockaddr_storage,
+    socklen_t, AF_INET, AF_INET6,
 };
+#[cfg(windows)]
+use winapi::shared::in6addr::IN6_ADDR as in6_addr;
+#[cfg(windows)]
+use winapi::shared::inaddr::IN_ADDR as in_addr;
 #[cfg(windows)]
 use winapi::shared::ws2def::{
     ADDRESS_FAMILY as sa_family_t, AF_INET, AF_INET6, SOCKADDR as sockaddr,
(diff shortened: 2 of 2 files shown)


Thomasdezeeuw:
I just review https://github.com/tokio-rs/mio/pull/1388, let's merge that first and than apply the same changes here. I'll also apply them to the master branch after.

Cargo.toml
```diff
@@ -21,7 +21,7 @@ features = ["handleapi", "ws2def", "ws2ipdef", "ws2tcpip", "minwindef"]
 
 [target."cfg(any(unix, target_os = \"redox\"))".dependencies]
 cfg-if = "0.1.6"
-libc = "0.2.66"
+libc = { version = "0.2.66", features = ["align"] }
```
faern, line 24: The added feature is so that `libc::sockaddr_in6` does not have an extra `__align` field (used before `#[repr(align(x))]` was a thing.)

faern:
What's currently in this PR now passes both CI and builds and runs all tests when built on top of a toolchain where the memory layout of `SocketAddrV4` and `SocketAddrV6` has changed. So I think this code is all right currently.

With the help of `[patch.crates-io]` I have also made sure `mio` builds and passes all tests when depending on this version of `socket2` and runs on my edited toolchain.

src/sockaddr.rs
```diff
@@ -168,34 +197,88 @@ impl SockAddr {
...
+            s_addr: u32::from_ne_bytes(addr.ip().octets()),
+        };
+        #[cfg(windows)]
+        let sin_addr = unsafe {
+            let mut s_un = mem::zeroed::<in_addr_S_un>();
+            *s_un.S_addr_mut() = u32::from_ne_bytes(addr.ip().octets());
+            in_addr { S_un: s_un }
+        };
+
+        let mut storage = MaybeUninit::<sockaddr_storage>::uninit();
```
faern, line 215: I'm thinking we should probably use `MaybeUninit::zeroed` to create these? Or possibly just `mem::zeroed()`. Currently after calling `assume_init` there are a lot of non-initialized bytes at the end of this `sockaddr_storage` struct.

faern: This file is currently a bit inconsistent. `SockAddr::from_raw_parts` use `MaybeUninit::<sockaddr_storage>::uninit()`, but `SockAddr::unix` uses `let mut addr = mem::zeroed::<sockaddr_un>();`

At least initializing all the structs to zero is probably a really good thing to do here. Whether or not directly with `mem::zeroed` or `MaybeUninit::zeroed` I don't have a strong opinion on.

src/sockaddr.rs
```diff
@@ -168,34 +199,88 @@ impl SockAddr {
...
+        let sockaddr_in = unsafe { &mut *(storage.as_mut_ptr() as *mut sockaddr_in) };
```
faern: I realized it's not very clean or safe to take out a safe Rust reference to the uninitialized memory region. A much better way is of course to use `storage.as_mut_ptr().write(value)` to guarantee that we just write a value into the region and don't use the reference for anything else that might be UB

faern:
I feel happy with the content of this PR now. Ready for review.

Thomasdezeeuw (approved):
It seems my previous comment didn't go through (I still see them as pending in the GitHub UI), so there might be double comments.

Thomasdezeeuw:
Thanks again @faern, one question left about the `align` feature of `libc`: is it also needed for Mio?

src/sockaddr.rs
Thomasdezeeuw: I had a comment that we likely needed to use zero: initialising only `sockaddr_in(6)` wasn't enough, as not the entire `sockaddr_storage` would be initialised.

src/sockaddr.rs
Thomasdezeeuw: Yes, the old code had UB. I had a comment about it but it's gone. The solution using `write` is perfect.

faern:
@Thomasdezeeuw I'm struggling to figure out when and where this is actually needed... Looks like my comment was a bit off, it's `in6_addr` that has an `__align` field on some platforms if this feature is not given. But a forgotten field would have given a compilation error, so it can't be needed in `mio`.

faern:
Looks like this feature is automatically activated if compiling on Rust 1.25 or newer: https://github.com/rust-lang/libc/blob/master/build.rs#L57-L59

So maybe it's because `socket2` has longer backwards compatibility? Does it?

faern:
So it might be needed for the `mio 0.6` backport? I'm not sure what the MSRV is for these projects, but it looks like `mio 0.7` [bumped it to 1.39](https://github.com/rust-lang/libc/blob/master/build.rs#L57-L59) so it should not be needed there! If latest `mio 0.6` is indeed still on [MSRV 1.18](https://github.com/tokio-rs/mio/blob/master/CHANGELOG.md#0612-january-5-2018) it will likely be needed for that backport.

For `socket2` I don't find any MSRV and there is no changelog. So I have no idea :open_mouth:

faern:
Wait... If `mio 0.6` has to support Rust older than 1.25 then we **can't** activate said feature. Because libc won't compile with that feature on too old Rust. So if we need to support so old Rust we need to provide the `__align` field. This is what I currently do with the help of `..unsafe { mem::zeroed() }` in my `net2` PR https://github.com/deprecrated/net2-rs/pull/106

PR #122, opened 2020-11-08
Don't assume memory layout of std::net::SocketAddr (master)

### LINKED ISSUE #121: Port pr #120 to master
Pr: #120
Issue: #119

faern:
Porting #120 from the 0.3.x branch to master.

Fixes #121

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -23,7 +23,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::Duration;
 use std::{cmp, fmt, io, mem};
 
-use libc::{self, c_void, ssize_t};
+use libc::{self, c_void, in6_addr, in_addr, ssize_t};
 
 #[cfg(not(target_os = "redox"))]
 use crate::RecvFlags;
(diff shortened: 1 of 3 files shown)


src/sys/unix.rs
```diff
@@ -1248,21 +1243,26 @@ fn timeval2dur(raw: libc::timeval) -> Option<Duration> {
...
+pub(crate) fn to_in_addr(addr: &Ipv4Addr) -> in_addr {
+    in_addr {
+        s_addr: u32::from_ne_bytes(addr.octets()),
+    }
 }
 
-fn from_s_addr(in_addr: libc::in_addr_t) -> Ipv4Addr {
-    in_addr.to_be().into()
+pub(crate) fn from_in_addr(in_addr: in_addr) -> Ipv4Addr {
+    Ipv4Addr::from(in_addr.s_addr.to_ne_bytes())
```
faern, line 1253: These helper methods were really handy. Removed a lot of code from the `sockaddr.rs` file and got rid of a number of `#[cfg(...)]`s. So I made them `pub(crate)` so I could use them in the other module. I alos changed the abstraction one level. So instead of converting to/from the only field in the `in_addr`/`in6_addr` I made these helpers convert directly into that struct. This in turn helped remove some repeated code here in the two sys files as well.

src/sys/unix.rs
```diff
@@ -1248,21 +1243,26 @@ fn timeval2dur(raw: libc::timeval) -> Option<Duration> {
...
-    let octets = addr.octets();
-    u32::from_ne_bytes(octets)
+pub(crate) fn to_in_addr(addr: &Ipv4Addr) -> in_addr {
+    in_addr {
+        s_addr: u32::from_ne_bytes(addr.octets()),
+    }
 }
 
-fn from_s_addr(in_addr: libc::in_addr_t) -> Ipv4Addr {
-    in_addr.to_be().into()
```
faern, line 1257: `.to_be()` here was not correct really... Well it yielded the same result, but it was not optimally from an understanding point of view IMO. `in_addr` is stored in BE. So `u32::from_be` would be more correct. But both of them just swap the bytes on a LE machine and does nothing on a BE machine, so it does not matter in practice.

Anyway. My solution uses `to_ne_bytes` to avoid a double swap (since the `From<u32>` impl for `Ipv4Addr` swaps the bytes back again).

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @faern.

PR #132, opened 2020-11-17
Additional Linux socket options

ammarzuberi:
Additional Linux socket options

Added support for setting/getting value of:
* TCP Maximum Segment Size (TCP_MAXSEG)
* Socket Mark (SO_MARK)

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -369,6 +369,49 @@ impl Socket {
         self.inner.set_ttl(ttl)
     }
 
+    /// Gets the value of the `TCP_MAXSEG` option on this socket.
+    ///
+    /// The `TCP_MAXSEG` option denotes the TCP Maximum Segment
+    /// Size and is only available on TCP sockets.
+    #[cfg(unix)]
+    pub fn mss(&self) -> io::Result<u32> {
+        self.inner.mss()
+    }
+
+    /// Sets the value of the `TCP_MAXSEG` option on this socket.
+    ///
+    /// The `TCP_MAXSEG` option denotes the TCP Maximum Segment
+    /// Size and is only available on TCP sockets.
+    #[cfg(unix)]
+    pub fn set_mss(&self, mss: u32) -> io::Result<()> {
+        self.inner.set_mss(mss)
+    }
+
+    /// Gets the value for the `SO_MARK` option on this socket.
+    ///
+    /// This value gets the socket mark field for each packet sent through
+    /// this socket.
+    ///
+    /// This function is only available on Linux and requires the
+    /// `CAP_NET_ADMIN` capability.
+    #[cfg(target_os = "linux")]
+    pub fn mark(&self) -> io::Result<u32> {
+        self.inner.mark()
+    }
+
+    /// Sets the value for the `SO_MARK` option on this socket.
+    ///
+    /// This value sets the socket mark field for each packet sent through
+    /// this socket. Changing the mark can be used for mark-based routing
+    /// without netfilter or for packet filtering.
+    ///
+    /// This function is only available on Linux and requires the
+    /// `CAP_NET_ADMIN` capability.
+    #[cfg(target_os = "linux")]
+    pub fn set_mark(&self, mark: u32) -> io::Result<()> {
+        self.inner.set_mark(mark)
+    }
+
     /// Gets the value of the `IPV6_UNICAST_HOPS` option for this socket.
     ///
     /// Specifies the hop limit for ipv6 unicast packets
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -534,6 +534,30 @@ impl Socket {
         unsafe { self.setsockopt(libc::IPPROTO_IP, libc::IP_TTL, ttl as c_int) }
     }
 
+    pub fn mss(&self) -> io::Result<u32> {
+        unsafe {
+            let raw: c_int = self.getsockopt(libc::IPPROTO_TCP, libc::TCP_MAXSEG)?;
+            Ok(raw as u32)
+        }
+    }
+
+    pub fn set_mss(&self, mss: u32) -> io::Result<()> {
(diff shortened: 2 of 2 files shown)


src/socket.rs
```diff
@@ -369,6 +369,49 @@ impl Socket {
...
     }
 
+    /// Gets the value for the `SO_MARK` option on this socket.
+    ///
+    /// This value gets the socket mark field for each packet sent through
+    /// this socket.
+    ///
+    /// This function is only available on Unix and requires the
+    /// `CAP_NET_ADMIN` capability.
+    #[cfg(target_os = "linux")]
```
de-vri-es, line 379: Hey. I'm not a maintainer so I can't really approve your PR. However, this crate also uses the `all` feature as an opt-in to platform specific functions.

So this should probably be:
```suggestion
    #[cfg(all(feature = "all", target_os = "linux"))]
```

You may want to apply that to all the new functions of your PR.

And, as a user of the crate, thank you for the contribution :)

src/socket.rs
```diff
@@ -369,6 +369,49 @@ impl Socket {
         self.inner.set_ttl(ttl)
     }
 
+    /// Gets the value for the `SO_MARK` option on this socket.
+    ///
+    /// This value gets the socket mark field for each packet sent through
+    /// this socket.
+    ///
+    /// This function is only available on Unix and requires the
```
Thomasdezeeuw, line 377: Linux only, not Unix only. Same for `set_mark`.

src/socket.rs
Thomasdezeeuw: As the maintainer, see above.

src/socket.rs
```diff
@@ -369,6 +369,49 @@ impl Socket {
...
+    #[cfg(target_os = "linux")]
+    pub fn set_mark(&self, mark: u32) -> io::Result<()> {
+        self.inner.set_mark(mark)
+    }
+
+    /// Gets the value of the `TCP_MAXSEG` option on this socket.
+    ///
+    /// The `TCP_MAXSEG` option denotes the TCP Maximum Segment
+    /// Size and is only available on TCP sockets on Linux platforms. 
+    #[cfg(target_os = "linux")]
```
Thomasdezeeuw, line 401: All Oses I've checked (FreeBSD, macOS, NetBSD, openBSD, Windows) have `TCP_MAXSEG`, please remove this `cfg`. Same for `set_mss`.

src/socket.rs
```diff
@@ -369,6 +369,49 @@ impl Socket {
...
 
+    /// Gets the value for the `SO_MARK` option on this socket.
+    ///
+    /// This value gets the socket mark field for each packet sent through
+    /// this socket.
+    ///
+    /// This function is only available on Unix and requires the
+    /// `CAP_NET_ADMIN` capability.
+    #[cfg(target_os = "linux")]
+    pub fn mark(&self) -> io::Result<u32> {
```
Thomasdezeeuw, line 380: Since this Unix (well Linux) only API please move it to `src/sys/unix.rs`, e.g. see https://github.com/rust-lang/socket2-rs/blob/dc82e662cda027f69abb75917777943fbffb8463/src/sys/unix.rs#L377-L417

src/socket.rs
Thomasdezeeuw: Although Windows only supports `TCP_MAXSEG` since Windows 10. Not sure what to do about that yet.

ammarzuberi: I dropped the `target_os = "Linux"` requirement.

src/sys/unix.rs
```diff
@@ -125,6 +125,19 @@ impl Type {
     pub fn cloexec(self) -> Type {
         Type(self.0 | libc::SOCK_CLOEXEC)
     }
+
+    #[cfg(all(feature = "all", target_os = "linux"))]
+    pub fn mark(&self) -> io::Result<u32> {
```
Thomasdezeeuw, line 130: Why did put this here?

src/socket.rs
```diff
@@ -369,6 +369,47 @@ impl Socket {
         self.inner.set_ttl(ttl)
     }
 
+    /// Gets the value for the `SO_MARK` option on this socket.
```
Thomasdezeeuw, line 372: https://github.com/rust-lang/socket2-rs/pull/132#discussion_r525316756 is not resolved. Remove the  `(set_)mark` functions here and move them to the `sys/unix.rs` file please. The same way `Socket::accept4` is exposed.

Thomasdezeeuw (changes requested):

src/sys/unix.rs
ammarzuberi: It was a PEBCAK issue.

src/socket.rs
ammarzuberi: I _think_ I've done that now. Let me know if it looks any better.

src/socket.rs
ammarzuberi: It looks like v0.3.x doesn't use the `all` feature anywhere.

src/socket.rs
```diff
@@ -369,6 +369,47 @@ impl Socket {
         self.inner.set_ttl(ttl)
     }
 
+    /// Gets the value of the `TCP_MAXSEG` option on this socket.
+    ///
+    /// The `TCP_MAXSEG` option denotes the TCP Maximum Segment
+    /// Size and is only available on TCP sockets.
+    pub fn mss(&self) -> io::Result<u32> {
```
Thomasdezeeuw, line 376: The CI fails on Windows, let's just mark it as unix only (`cfg[unix]`) for now and we'll figure out Windows 10 support when someone requests it.

Thomasdezeeuw (approved):
I was thinking this pr was against the master branch, which has a different structure to v0.3. I've marked most conversation as resolved because of it.

Left one last comment, LGTM otherwise.

ammarzuberi:
It should be ready to be merged now. Is there a regular release cadence for v0.3.x? If not, could you please cut a release?

Thomasdezeeuw:
I'll make a release this weekend.

PR #137, opened 2020-11-29
Fix compilation on Haiku on 0.3.x

linkmauve:
`sin_len` and `sin6_len` are apparently required on all BSD-like systems, from which Haiku probably inherits its network stack.

`sin_zero` has been changed from 8 bytes to 24 bytes in hrev9079 in order to stay compatible with BeOS.

This was apparently broken in #120.

Same as #136 but for the v0.3.x branch.


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @linkmauve.

PR #140, opened 2020-12-08
Enable some constants that are actually available on redox

coolreader18:
`libc` recently added more constants for redox, so some of the cfgs can be removed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,9 +33,9 @@ features = ["all"]
 version = "0.3.3"
 features = ["handleapi", "ws2def", "ws2ipdef", "ws2tcpip", "minwindef"]
 
-[target."cfg(any(unix, target_os = \"redox\"))".dependencies]
+[target."cfg(unix)".dependencies]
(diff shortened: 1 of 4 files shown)


Thomasdezeeuw (commented):
You've remove the `all` feature on a lot constants and functions.

```
#[cfg(all(feature = "all", not(target_os = "redox")))]
```
Should become:
```
#[cfg(feature = "all")]
```

coolreader18:
Ohh, I figured that those were only there because redox didn't have them, but I assume windows doesn't either? I'll fix that

coolreader18:
I don't think the macos failure is related to my changes

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @coolreader18. So you know, this is the v0.4.x version which is currently released and doesn't have a release date yet.

coolreader18:
Yup, I was going to backport these redox fixes to 0.3 next, as it currently doesn't compile.

PR #144, opened 2020-12-14
add bind to device function

kolapapa:
Bind a socket to a particular device like "eth0", as specified in the passed interface name. 

In some cases, we hope that the socket can be directly bound to the specified interface. All along, after the socket is created with `socket2`, we manually use `setsockopt` to bind. We hope to support this function on the official library.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -867,6 +867,72 @@ impl crate::Socket {
         unsafe { setsockopt::<c_int>(self.inner, libc::SOL_SOCKET, libc::SO_MARK, mark as c_int) }
     }
 
+    /// Gets the value for the `SO_BINDTODEVICE` option on this socket.
+    ///
+    /// This value gets the socket binded device's interface name.
+    ///
+    /// This function is only available on Linux.
+    #[cfg(all(feature = "all", target_os = "linux"))]
+    pub fn device(&self) -> io::Result<Option<String>> {
+        unsafe {
+            let mut opt_len = libc::IFNAMSIZ as libc::socklen_t;
+            let mut payload: MaybeUninit<[u8; libc::IFNAMSIZ]> = MaybeUninit::uninit();
+            syscall!(getsockopt(
+                self.inner,
+                libc::SOL_SOCKET,
+                libc::SO_BINDTODEVICE,
+                payload.as_mut_ptr().cast(),
+                &mut opt_len,
+            ))?;
+            match opt_len {
+                0 => Ok(None),
+                _ => {
+                    let end: usize = opt_len as usize - 1;
+                    match String::from_utf8(payload.assume_init()[..end].to_vec()) {
+                        Ok(interface) => Ok(Some(interface)),
+                        Err(_) => {
+                            return Err(io::Error::new(
+                                io::ErrorKind::InvalidData,
+                                "Data not valid for the operation were encountered.",
+                            ));
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    /// Sets the value for the `SO_BINDTODEVICE` option on this socket.
+    ///
+    /// If a socket is bound to an interface, only packets received from that
+    /// particular interface are processed by the socket. Note that this only
+    /// works for some socket types, particularly AF_INET sockets. Accept an
+    /// Option where None which will set the length to zero, removing the
+    /// interface binding.
+    ///
+    /// This function is only available on Linux.
+    #[cfg(all(feature = "all", target_os = "linux"))]
+    pub fn bind_device(&self, interface: Option<&str>) -> io::Result<()> {
+        unsafe {
+            let value = interface.unwrap_or_default();
+            let opt_len = value.len();
+            if opt_len > libc::IFNAMSIZ {
+                return Err(io::Error::new(
+                    io::ErrorKind::InvalidInput,
+                    "interface name is too long",
+                ));
+            }
+            syscall!(setsockopt(
+                self.inner,
+                libc::SOL_SOCKET,
+                libc::SO_BINDTODEVICE,
+                value.as_ptr() as *const c_void,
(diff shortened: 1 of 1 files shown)


src/socket.rs
```diff
@@ -956,6 +958,18 @@ impl Socket {
...
 
+    /// Set value for the `SO_BINDTODEVICE` option on this socket.
+    ///
+    /// If a socket is bound to an interface, only packets received from that
+    /// particular interface are processed by the socket. Note that this only
+    /// works for some socket types, particularly AF_INET sockets.
+    ///
+    /// This function is only available on Linux.
+    #[cfg(all(feature = "all", target_os = "linux"))]
+    pub fn set_bind_interface(&self, interface: CString) -> io::Result<()> {
```
Thomasdezeeuw, line 969: `CString` is an owned version, since we just need to use for a while, so we can use `CStr` (the borrowed version) instread.

Thomasdezeeuw: This should accept an `Option` where `None` which will set the length to zero, removing the interface binding. Or we could add a `remove_bind_interface` function which does this.

src/sys/unix.rs
```diff
@@ -948,6 +950,20 @@ impl Socket {
...
 
+    #[cfg(all(feature = "all", target_os = "linux"))]
+    pub fn set_bind_interface(&self, interface: CString) -> io::Result<()> {
+        unsafe {
+            syscall!(setsockopt(
+                self.fd,
+                libc::SOL_SOCKET,
+                libc::SO_BINDTODEVICE,
+                interface.as_ptr() as *const c_void,
+                mem::size_of::<CString>() as libc::socket_t,
```
Thomasdezeeuw, line 961: Shouldn't this be the length of the string?

src/socket.rs
Thomasdezeeuw, line 969: Should be name `bind_device` to match the `SO_BINDTODEVICE` constant name?

Thomasdezeeuw, line 969: We should also have a getter, e.g. `device(&self) -> Option<CString>`. Or allow the user to pass a `CString`.

Thomasdezeeuw, line 969: Since this a Linux (Unix?) only option it should be moved to the `sys/unix.rs` file, for an example of this see the `Socket::set_mark` function: https://github.com/rust-lang/socket2-rs/blob/be524550f6cfb5557637cb56bbba6e62f2ff5a89/src/sys/unix.rs#L686-L698

Thomasdezeeuw (commented):
Left a couple of comments, but overall I'm good with the addition.

You should know that this chance will go into the v0.4 release, which doesn't have a release date yet. If you want it to go into the v0.3.x you can open a pr against the v0.3.x branch.

kolapapa:
Thank you for your detailed comments. I have benefited a lot from it. I have submitted a new edition. Please take time to review it.

kolapapa:
> You should know that this chance will go into the v0.4 release, which doesn't have a release date yet. If you want it to go into the v0.3.x you can open a pr against the v0.3.x branch.

OK, when this PR is merged, I will open another one~

kolapapa:
Is there any plan to review this PR? Or if there is a problem, I will change it?

Thomasdezeeuw:
@kolapapa please take a look at the CI, because it's failing see https://github.com/rust-lang/socket2-rs/pull/144/checks?check_run_id=1581114191.

kolapapa:
> @kolapapa please take a look at the CI, because it's failing see https://github.com/rust-lang/socket2-rs/pull/144/checks?check_run_id=1581114191.

Is there anything else I need to modify?

src/sys/unix.rs
```diff
@@ -867,6 +867,72 @@ impl crate::Socket {
...
+                libc::SOL_SOCKET,
+                libc::SO_BINDTODEVICE,
+                payload.as_mut_ptr().cast(),
+                &mut opt_len,
+            ))?;
+            match opt_len {
+                0 => Ok(None),
+                _ => {
+                    let end: usize = opt_len as usize - 1;
+                    match String::from_utf8(payload.assume_init()[..end].to_vec()) {
```
Thomasdezeeuw, line 891: This is undefined behaviour. With `assume_init` your saying that `payload` is fully initialised, but only `opt_len` bytes are.

Thomasdezeeuw:
@kolapapa could you check if https://github.com/Thomasdezeeuw/socket2/commit/3851430dec41c204d6219a84c47ca6885622a98e works for you?

kolapapa:
> Thomasdezeeuw@3851430

It works~

I learned from your commit that the code block of unsafe should be as short as possible. 
However, I have a question: 
why must `CString` or `Cstr` be used here? I think `String` or `str` is more convenient to call. It doesn't have to be `CString` or `Cstr`?

Thomasdezeeuw:
> It works~
> 
> I learned from your commit that the code block of unsafe should be as short as possible.

Yes that is usually desirable, also a comment prefixed with `// Safety:` explaining why the call is actually safe.

> However, I have a question:
> why must `CString` or `Cstr` be used here? I think `String` or `str` is more convenient to call. It doesn't have to be `CString` or `Cstr`?

`String` and `str` must be valid UTF-8, however in the documentation of the system call I couldn't find any guarantees that the device name has to be UTF-8. To stay as flexible also possible it's best to use `CString`/`CStr` in those cases which are specially designed for using with C code.

src/sys/unix.rs
kolapapa: Thank you for your advice. I haven't been proficient in using `slice` to interact...

kolapapa:
@Thomasdezeeuw  Thanks a lot, I have gained a lot in this submission. Now I will close this PR and you can merge your commit?

Thomasdezeeuw:
Done in #155.

Issue #78, opened 2020-05-07
Tracking issue: expand CI testing


Thomasdezeeuw:
CI support, following the same tiers as rust (https://doc.rust-lang.org/nightly/rustc/platform-support.html).

## Tier 1

Tests must pass.

 * [x] Linux: tested.
 * [x] Windows: tested.
 * [x] macOS: tested.

## Tier 2

Must build.

 * [x] Android: check only.
 * [x] FreeBSD: check only.
 * [x] Fuchsia: check only.
 * [x] NetBSD: check only.
 * [x] Redox: check using Rust nightly only, stable is tracked here: https://github.com/rust-lang/socket2/issues/181.
 * [x] Solaris: check only.
 * [x] iOS: check only.
 * [x] illumos: check only.

## Tier 3

Best effort.

 * [ ] Haiku
 * [ ] Hermit
 * [ ] OpenBSD
 * [ ] VxWorks
 * [ ] DragonFly BSD

<hr>

Old description:

Currently only Linux, macOS and Windows are tested on the CI but we support (to some degree) many more OSes. We should add more of these OSes to the CI, at least running `cargo check`. For (Rust) tier 1 architecture/OS combos this should be easy, but for other tiers `rustup` might not be supported.

KodrAus:
cc @rust-lang/infra I think we have a few repos in `rust-lang` with fairly extensive CI suites. Can we think of any examples that we could use to bootstrap some broader CI here? I'm thinking particularly of iOS...

Thomasdezeeuw:
@KodrAus, @rust-lang/infra a little update: tier 2 targets are now all `cargo check`ed, but if it's possible to actually run the tests somehow I would prefer that.

KodrAus:
@Thomasdezeeuw I'm guessing it's more the OS than the architecture that you're interested in for the tier 2 targets?

It looks like we might be able to [use MacOS workers in GitHub Actions to run some of the BSDs](https://github.com/vmactions/freebsd-vm). Alternatively, Travis CI does seem to support FreeBSD (although I can't find any docs on it), which we do still use in some repos here in `rust-lang` (specifically `stdarch` and `stdsimd`).

Thomasdezeeuw:
> @Thomasdezeeuw I'm guessing it's more the OS than the architecture that you're interested in for the tier 2 targets?

[...]

PR #154, opened 2020-12-20
Add links to OS documentation

Thomasdezeeuw:
Depends on extended_key_value_attributes, tracked here:
https://github.com/rust-lang/rust/issues/78835.

Creates the following links:

 * DragonFly BSD: https://man.dragonflybsd.org/?command=recv&section=2
 * FreeBSD: https://www.freebsd.org/cgi/man.cgi?query=recv&sektion=2
 * illumos: https://illumos.org/man/3SOCKET/recv
 * Linux: https://man7.org/linux/man-pages/man2/recv.2.html
 * NetBSD: https://man.netbsd.org/recv.2
 * OpenBSD: https://man.openbsd.org/recv.2
 * Windows: https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-recv

PR #164, opened 2020-12-24
Add Fuchsia support and CI

tamird:
In general, feature support is identical to Linux.

Updates #78.


src/sys/unix.rs
```diff
@@ -519,15 +526,14 @@ fn recvmsg(
...
-    let mut msg = libc::msghdr {
-        msg_name: msg_name.cast(),
-        msg_namelen,
-        msg_iov: bufs.as_mut_ptr().cast(),
-        msg_iovlen: min(bufs.len(), IovLen::MAX as usize) as IovLen,
-        msg_control: ptr::null_mut(),
-        msg_controllen: 0,
-        msg_flags: 0,
-    };
+    let mut msg: libc::msghdr = unsafe { mem::zeroed() };
```
Thomasdezeeuw, line 529: Can you revert back to creating the structure directly, it makes a bit more awkward with differences between OSs I know, but it forces you to check the structure when porting to a new OS (like what happend here).

src/sys/unix.rs
```diff
@@ -578,19 +584,18 @@ fn sendmsg(
...
-        msg_name: (msg_name as *mut sockaddr_storage).cast(),
-        msg_namelen,
-        // Safety: Same as above about `*const` -> `*mut`.
-        msg_iov: bufs.as_ptr() as *mut _,
-        msg_iovlen: min(bufs.len(), IovLen::MAX as usize) as IovLen,
-        msg_control: ptr::null_mut(),
-        msg_controllen: 0,
-        msg_flags: 0,
-    };
+    let mut msg: libc::msghdr = unsafe { mem::zeroed() };
```
Thomasdezeeuw, line 587: Same as above.

Thomasdezeeuw (approved):
Two small things, but LGTM otherwise.

Thomasdezeeuw:
@tamird is there manual online for libc/POSIX complicated system call interface for Fuchsia? A link I can add to https://github.com/rust-lang/socket2/pull/154?

Thomasdezeeuw:
For myself, update https://github.com/rust-lang/socket2/issues/78.

src/sys/unix.rs
tamird: The trouble is that this structure has private padding fields on Fuchsia, and can't be instantiated directly.

Shall I add a comment? Is there a better solution?

tamird:
> @tamird is there manual online for libc/POSIX complicated system call interface for Fuchsia? A link I can add to #154?

Not that I'm aware of at this time. Fuchsia's libc is derived from musl, so it is largely in line with Linux (or is intended to be, in areas where it is not yet so).

src/sys/unix.rs
Thomasdezeeuw: Does something like the following work?
```
msghdr {
    // fields ...
    .. Default::default(),
}
```

Or does `msghdr` need to implement `Default` for that? I forgot...

If it's not possible, keep the change as is but with a comment indeed. And can you remove the `msg_controllen` and `msg_flags` fields as there already zero.

tamird: Yeah, it has to impl `Default`. Added a comment and removed zero initialization.

src/sys/unix.rs
tamird: Done.

Thomasdezeeuw:
Thanks @tamird.

Thomasdezeeuw:
Hey @tamird, apparently Fuchsia doesn't yet work. I tried checking it locally and found the following errors:

```log
error[E0425]: cannot find value `IFNAMSIZ` in crate `libc`
   --> src/sys/unix.rs:934:46
    |
934 |         let mut buf: [MaybeUninit<u8>; libc::IFNAMSIZ] =
    |                                              ^^^^^^^^ not found in `libc`

error[E0425]: cannot find value `IFNAMSIZ` in crate `libc`
   --> src/sys/unix.rs:935:60
    |
935 |             unsafe { MaybeUninit::<[MaybeUninit<u8>; libc::IFNAMSIZ]>::uninit().assume_init() };
    |                                                            ^^^^^^^^ not found in `libc`

error[E0425]: cannot find value `IP_FREEBIND` in crate `libc`
    --> src/sys/unix.rs:1049:69
     |
1049 |             getsockopt::<c_int>(self.inner, libc::SOL_SOCKET, libc::IP_FREEBIND)
     |                                                                     ^^^^^^^^^^^ not found in `libc`

error[E0425]: cannot find value `IP_FREEBIND` in crate `libc`
    --> src/sys/unix.rs:1072:23
     |
1072 |                 libc::IP_FREEBIND,
     |                       ^^^^^^^^^^^ not found in `libc`

```

So `IFNAMSIZ` and `IP_FREEBIND` aren't defined. It's also a problem in the CI, but it passed anyway (now issue #169). Are these two constant available already in Fuchsia?

tamird:
That's odd! Yeah, those constants are defined in our libc:

`IFNAMSIZ`: https://cs.opensource.google/fuchsia/fuchsia/+/master:zircon/third_party/ulib/musl/include/net/if.h;l=72;drc=bdf60afa9193c8a9190ee6bb6b5a0a11ad8876b6

`IP_FREEBIND`: https://cs.opensource.google/fuchsia/fuchsia/+/master:zircon/third_party/ulib/musl/include/netinet/in.h;l=196;drc=bdf60afa9193c8a9190ee6bb6b5a0a11ad8876b6

but apparently not in rust-lang/libc:

`IFNAMSIZ` should be here: https://github.com/rust-lang/libc/blob/c9fb51bc650427fe22dcf34b503f6aa6fc909c07/src/fuchsia/mod.rs#L1329 -- sent https://github.com/rust-lang/libc/pull/2010.

`IP_FREEBIND` should be here: https://github.com/rust-lang/libc/blob/c9fb51bc650427fe22dcf34b503f6aa6fc909c07/src/fuchsia/mod.rs#L1765 -- sent https://github.com/rust-lang/libc/pull/2011.

PR #186, opened 2020-12-29
Remove FFI types from the public API

tamird:
Extracted from #184 @Thomasdezeeuw.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -7,11 +7,6 @@
 // except according to those terms.
 
 use std::cmp::min;
-#[cfg(all(
-    feature = "all",
-    any(target_os = "android", target_os = "fuchsia", target_os = "linux")
-))]
-use std::ffi::{CStr, CString};
 #[cfg(not(target_os = "redox"))]
 use std::io::IoSlice;
 use std::marker::PhantomData;
@@ -942,10 +937,10 @@ impl crate::Socket {
         feature = "all",
         any(target_os = "android", target_os = "fuchsia", target_os = "linux")
     ))]
-    pub fn device(&self) -> io::Result<Option<CString>> {
+    pub fn device(&self) -> io::Result<Option<Vec<u8>>> {
         // TODO: replace with `MaybeUninit::uninit_array` once stable.
         let mut buf: [MaybeUninit<u8>; libc::IFNAMSIZ] =
-            unsafe { MaybeUninit::<[MaybeUninit<u8>; libc::IFNAMSIZ]>::uninit().assume_init() };
+            unsafe { MaybeUninit::uninit().assume_init() };
         let mut len = buf.len() as libc::socklen_t;
         unsafe {
             syscall!(getsockopt(
@@ -959,21 +954,9 @@ impl crate::Socket {
         if len == 0 {
             Ok(None)
         } else {
-            // Allocate a buffer for `CString` with the length including the
-            // null terminator.
-            let len = len as usize;
-            let mut name = Vec::with_capacity(len);
-
+            let buf = &buf[..len as usize - 1];
             // TODO: use `MaybeUninit::slice_assume_init_ref` once stable.
-            // Safety: `len` bytes are writen by the OS, this includes a null
-            // terminator. However we don't copy the null terminator because
-            // `CString::from_vec_unchecked` adds its own null terminator.
-            let buf = unsafe { slice::from_raw_parts(buf.as_ptr().cast(), len - 1) };
-            name.extend_from_slice(buf);
-
-            // Safety: the OS initialised the string for us, which shouldn't
-            // include any null bytes.
-            Ok(Some(unsafe { CString::from_vec_unchecked(name) }))
+            Ok(Some(unsafe { &*(buf as *const [_] as *const [u8]) }.into()))
         }
     }
 
@@ -990,9 +973,9 @@ impl crate::Socket {
         feature = "all",
         any(target_os = "android", target_os = "fuchsia", target_os = "linux")
     ))]
-    pub fn bind_device(&self, interface: Option<&CStr>) -> io::Result<()> {
+    pub fn bind_device(&self, interface: Option<&[u8]>) -> io::Result<()> {
         let (value, len) = if let Some(interface) = interface {
-            (interface.as_ptr(), interface.to_bytes_with_nul().len())
+            (interface.as_ptr(), interface.len())
         } else {
             (ptr::null(), 0)
         };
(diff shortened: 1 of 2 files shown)


src/sys/unix.rs
```diff
@@ -959,21 +956,9 @@ impl crate::Socket {
...
-            // terminator. However we don't copy the null terminator because
-            // `CString::from_vec_unchecked` adds its own null terminator.
-            let buf = unsafe { slice::from_raw_parts(buf.as_ptr().cast(), len - 1) };
-            name.extend_from_slice(buf);
-
-            // Safety: the OS initialised the string for us, which shouldn't
-            // include any null bytes.
-            Ok(Some(unsafe { CString::from_vec_unchecked(name) }))
+            let buf = &buf[..len.try_into().unwrap()];
+            Ok(Some(unsafe { mem::transmute::<_, &[u8]>(buf) }.into()))
```
Thomasdezeeuw, line 961: Let's not use `transmute`, it's too easy to misuse. Also let's keep the same code as in `MaybeUninit::slice_assume_init_ref` so we can switch to that easily once stable.

It think the following would be shortest.

```rust

// Safety: `len` bytes are writen by the OS, this includes a null
// terminator. However we don't need the null terminator.
let buf = unsafe { slice::from_raw_parts(buf.as_ptr().cast(), (len - 1) as usize) };
Ok(Some(buf.into_vec()))
```

Thomasdezeeuw (commented):
After discussing this in https://github.com/rust-lang/socket2/pull/184#discussion_r549661768, I'm ok with the change.

src/sys/unix.rs
```diff
@@ -942,11 +937,13 @@ impl crate::Socket {
         feature = "all",
         any(target_os = "android", target_os = "fuchsia", target_os = "linux")
     ))]
-    pub fn device(&self) -> io::Result<Option<CString>> {
+    pub fn device(&self) -> io::Result<Option<Vec<u8>>> {
+        use std::convert::TryInto as _;
```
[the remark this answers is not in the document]
tamird: Ack. I usually do this with traits to avoid bringing the trait name into the local namespace.

src/sys/unix.rs
tamird: `len` does _not_ include a null terminator AFAIK.

Other than that, this code _is_ the same as it would be with `slice_assume_init_ref`. That code would be:
```rust
let buf = &buf[..len.try_into().unwrap()];
Ok(Some(unsafe { MaybeUninit::slice_assume_init_ref(buf) }.into()))
```

src/sys/unix.rs
```diff
@@ -942,11 +937,13 @@ impl crate::Socket {
...
+
         // TODO: replace with `MaybeUninit::uninit_array` once stable.
         let mut buf: [MaybeUninit<u8>; libc::IFNAMSIZ] =
-            unsafe { MaybeUninit::<[MaybeUninit<u8>; libc::IFNAMSIZ]>::uninit().assume_init() };
-        let mut len = buf.len() as libc::socklen_t;
+            unsafe { MaybeUninit::uninit().assume_init() };
+        let mut len = buf.len().try_into().unwrap();
```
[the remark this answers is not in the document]
tamird: The buf initialization _is_ the same as the original code (in this PR, not in #184); I've just removed some repeated type parameters.

I agree that IFNAMSIZ in practice can't fail this conversion, but why think about it? This code is safer and should compile to the same thing.

src/sys/unix.rs
Thomasdezeeuw: `len` does include a null terminate, I've just manually tried this patch. The test currently doesn't fail it it doesn't have permissions (mainly for the CI), but we might have to change that. I'll fiddle with this a bit more and merge this changing the signature to `[u8]` rather than `CStr`.

P.S. it requires `sudo setcap cap_net_raw=ep $test_bin`, something that we might have to do in the CI as well.

Thomasdezeeuw:
@tamird created #187 which works. I've put `ignore` attribute to the test as it doesn't run without the `CAP_NET_RAW` capability (on Linux).

src/sys/unix.rs
tamird: I'm about 95% sure that len *does not* include the null terminator when the name is IFNAMSIZ characters long. That is probably an edge case worth testing.

tamird:
@Thomasdezeeuw you should be able to push to this PR rather than working in a separate one.

Thomasdezeeuw:
@tamird I didn't know I could push to some else's branch.

tamird:
@Thomasdezeeuw https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

src/sys/unix.rs
Thomasdezeeuw: After a bit more manual testing. I couldn't create an interface with a name larger than 15 bytes, with `IFNAMSIZ` being 16.

```bash
sudo ip link add test_interface0 type bridge  # Fine.
sudo ip link add test_interface01 type bridge # Error: Attribute failed policy validation.
```

I also tested binding to `test_interface01` with the implementation in #187 which worked.

Thomasdezeeuw:
@tamird I've pushed the changes to this branch.

tests/socket.rs
```diff
@@ -631,30 +629,40 @@ fn tcp_keepalive() {
...
 
     let socket = Socket::new(Domain::IPV4, Type::STREAM, None).unwrap();
     assert_eq!(socket.device().unwrap(), None);
 
     for interface in INTERFACES.iter() {
-        let interface = CStr::from_bytes_with_nul(interface.as_bytes()).unwrap();
-        if let Err(err) = socket.bind_device(Some(interface)) {
+        if let Err(err) = socket.bind_device(Some(interface.as_bytes())) {
             // Network interface is not available try another.
             if matches!(err.raw_os_error(), Some(libc::ENODEV) | Some(libc::EPERM)) {
```
tamird, line 643: We should stop allowing EPERM now, yes?

tests/socket.rs
```diff
@@ -631,30 +629,40 @@ fn tcp_keepalive() {
...
         assert_eq!(socket.device().unwrap(), None);
         // Just need to do it with one interface.
-        break;
+        return;
     }
+
+    panic!(
+        "failing to bind to any device. \
+        Note that on Linux this requires the `CAP_NET_RAW` permission. \
+        This can be given using `sudo setcap cap_net_raw=ep $test_bin`."
```
tamird, line 664: Should this mention whatever flag is necessary to run this test case given the ignore attribute?

Thomasdezeeuw: I've changed it to just say "failed to bind to any device", the other info is already in the ignore attribute.

With flag do you mean `--ignore` passed to `cargo test`, or do mean another flag? If the first than I don't it's needed.

PR #194, opened 2021-01-08
Add back `Socket::connect_timeout`

### LINKED ISSUE #190: Support for connect_timeout in 0.4
As I saw in 0.4 changelogs `Socket::connect_timeout` was removed because it was not a low-level function but a combination of system calls. This is true, however a lot of people may need this functionality.

For someone to implement this and make it cross-platform, it will be troublesome. So can we have it back? Maybe we can have it in a new crate? For example in `socket2-util` crate and in `SocketExt` trait.

oblique:
Fixes #190


Thomasdezeeuw:
@oblique I've pushed a commit (twice) to reduce the diff.

Thomasdezeeuw:
Thanks @oblique.

PR #195, opened 2021-01-11
Add some test support for BSDs

KodrAus:
Part of #78 

This PR attempts to use https://github.com/vmactions/freebsd-vm to run tests in CI on a FreeBSD target. It appears to use VirtualBox virtualization on a MacOS host.

If I manage to get things working on this target then I'll try NetBSD and Solaris as well.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,7 +58,7 @@ jobs:
       matrix:
         # TODO: add the following targets, currently broken.
         # "x86_64-unknown-redox"
-        target: ["aarch64-apple-ios", "aarch64-linux-android", "x86_64-apple-darwin", "x86_64-fuchsia", "x86_64-pc-windows-msvc", "x86_64-sun-solaris", "x86_64-unknown-freebsd", "x86_64-unknown-illumos", "x86_64-unknown-linux-gnu", "x86_64-unknown-netbsd"]
+        target: ["aarch64-apple-ios", "aarch64-linux-android", "x86_64-apple-darwin", "x86_64-fuchsia", "x86_64-pc-windows-msvc", "x86_64-sun-solaris", "x86_64-unknown-illumos", "x86_64-unknown-linux-gnu", "x86_64-unknown-netbsd"]
     steps:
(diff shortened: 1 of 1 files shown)


KodrAus:
Looks like we've got some tests running in FreeBSD along with a failure: https://github.com/rust-lang/socket2/runs/1679158951?check_suite_focus=true#step:3:1567

```
failures:

---- only_v6 stdout ----
thread 'main' panicked at 'assertion failed: `(left != right)`
  left: `true`,
 right: `true`: initial value and argument are the same', tests/socket.rs:772:1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Thomasdezeeuw:
@KodrAus fixed the test issue in #196.

.github/workflows/main.yml
```diff
@@ -89,6 +89,26 @@ jobs:
...
+      matrix:
+        target: ["x86_64-unknown-freebsd"]
+    steps:
+    - uses: actions/checkout@master
+    - name: Run check
+      uses: vmactions/freebsd-vm@v0.1.0
+      with:
+        prepare: |
+          pkg install -y curl
+          curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly
```
Thomasdezeeuw, line 105: Two small things. Can you switch this to stable?

And can you add `--profile minimal`, it should speed up the download (a bit).

Thomasdezeeuw (approved):
Two small things, but LGTM otherwise. Can you rebase this on master to check if I fixed the ipv6_only test failure with #196.

KodrAus:
Done! Let's see if it passes now.

You may also be interested in https://github.com/vmactions/freebsd-vm/issues/8, which explains why the FreeBSD output contains a lot of garbled text right now before it gets to our code.

KodrAus:
Looks like we're green now.

If you like I should be able to follow this one up with NetBSD and Solaris too.

Thomasdezeeuw:
@KodrAus thanks. Adding NetBSD and Solaris would be great.

PR #198, opened 2021-01-11
Add AF_VSOCK support

elmarco:
"The VSOCK address family facilitates communication between virtual machines and the host they are running on.  This address family is used by guest agents and hypervisor services that need a communications channel that is in‐dependent of virtual machine network configuration." (vsock(7))

It is currently Linux-only here, but it should also be available for other systems, Windows and BSD in particular.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -463,8 +463,20 @@ impl SockAddr {
         }
         .map(|(_, addr)| addr)
     }
-}
 
+    /// Returns this address VSOCK CID/port if it is in the `AF_VSOCK` family,
+    /// otherwise return `None`.
+    #[cfg(all(feature = "all", any(target_os = "android", target_os = "linux")))]
+    pub fn vsock_address(&self) -> Option<(u32, u32)> {
+        if self.family() == libc::AF_VSOCK as sa_family_t {
+            // Safety: if the ss_family field is AF_VSOCK then storage must be a sockaddr_vm.
+            let addr = unsafe { &*(self.as_ptr() as *const libc::sockaddr_vm) };
+            Some((addr.svm_cid, addr.svm_port))
+        } else {
+            None
+        }
+    }
+}
 
 pub(crate) type Socket = c_int;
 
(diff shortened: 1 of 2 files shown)


src/sys/unix.rs
```diff
@@ -406,6 +416,46 @@ impl SockAddr {
...
+                Ok(())
+            })
+        }
+        .map(|(_, addr)| addr)
+    }
+
+    /// Returns this address VSOCK port if it is in the `AF_VSOCK` family,
+    /// otherwise return `None`.
+    #[cfg(all(feature = "all", any(target_os = "android", target_os = "linux")))]
+    pub fn vsock_port(&self) -> Option<u32> {
```
Thomasdezeeuw, line 448: Is the Context id interesting at all? I.e. it is worth returning (cid, port) (`(u32, u32)`) here?

tests/socket.rs
```diff
@@ -112,6 +114,14 @@ fn socket_address_unix() {
     assert!(addr.as_socket_ipv6().is_none());
 }
 
+#[test]
+#[cfg(all(feature = "all", any(target_os = "android", target_os = "linux")))]
+fn socket_address_vsock() {
+    let addr = SockAddr::vsock(1, 9999).unwrap();
+    assert!(addr.as_socket_ipv4().is_none());
+    assert!(addr.as_socket_ipv6().is_none());
```
Thomasdezeeuw, line 122: Can you add `assert!(addr.vsock_port(), Some(9999));`

Thomasdezeeuw (approved):
I don't use this my self, except for two small things the code looks good to me.

We can add more OS support in the follow-up prs.

src/sys/unix.rs
elmarco: Yes, I should have done so, my bad.

tests/socket.rs
elmarco: done, thanks for the suggestion

elmarco:
Running the tests in CI seems problematic. Should we mark it ignored?

Thomasdezeeuw:
> Running the tests in CI seems problematic. Should we mark it ignored?

Yes, but please clearly describe what is required to run the test in the ignore reason, e.g. https://github.com/rust-lang/socket2/blob/4130e75c6cb36041839991a26b22c5145cc19d15/tests/socket.rs#L701.

Thomasdezeeuw:
Thanks @elmarco.

elmarco:
yw, thanks!

PR #206, opened 2021-02-01
Fuchsia doesn't support sendfile

erickt:
Fuchsia doesn't support `sendfile`.

edit: previously I asked for a point release with this because of the typo in "fuchsia", but we don't actually need it since this was for an unsupported syscall.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -18,7 +18,6 @@ use std::net::{Ipv4Addr, Ipv6Addr};
     any(
         target_os = "android",
         target_os = "freebsd",
-        target_os = "fucsia",
         target_os = "linux",
         target_vendor = "apple",
     )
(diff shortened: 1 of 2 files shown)


tamird:
Looks like some of this fails to compile; we should just remove those parts.

```
error[E0425]: cannot find function `sendfile` in crate `libc`
    --> src/sys/unix.rs:1384:18
     |
1384 |         syscall!(sendfile(self.inner, file, &mut offset, count)).map(|n| n as usize)
     |                  ^^^^^^^^ not found in `libc`

error[E0425]: cannot find function `sendfile` in crate `libc`
    --> src/sys/unix.rs:1384:18
     |
1384 |         syscall!(sendfile(self.inner, file, &mut offset, count)).map(|n| n as usize)
     |                  ^^^^^^^^ not found in `libc`
```

erickt:
Turns out fuchsia doesn't support `sendfile`. I just removed these lines instead.

Thomasdezeeuw (approved):

Thomasdezeeuw:
@erickt can you run rustfmt on this, then I think it's good to go.

Thomasdezeeuw:
Thanks @erickt

PR #214, opened 2021-03-19
Fix breaking tests due to libc's deprecation of Musl time_t

oherrala:
Musl 1.2 has switched to 64 bit `time_t`, and `libc` crate is giving deprecation warning about it on Musl targets. This PR silences that this warnings so that it doesn't break tests.

libc issue: https://github.com/rust-lang/libc/issues/1848

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -769,6 +769,8 @@ pub(crate) fn set_timeout_opt(
 
 fn into_timeval(duration: Option<Duration>) -> libc::timeval {
     match duration {
+        // https://github.com/rust-lang/libc/issues/1848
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
If I've read the issues/prs correctly `timt_t` will change from `i32` (on 32 bit) to `i64` in musl 1.2, right? Since we're already capping the value the code should continue to work with that change. Then I'm +1.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @oherrala.

oherrala:
@Thomasdezeeuw yes, I think `socket2`'s code is not affected by Musl 1.2 change. And the deprecation warning hopefully goes away someday. Thanks for the quick merge! 👍

PR #217, opened 2021-03-25
Fix endianness issues in sys::unix::in_addr_convertion test

### LINKED ISSUE #216: 0.4.0: sys::in_addr_convertion test fails on s390x / IBM System Z (big-endian)
```
---- sys::in_addr_convertion stdout ----
thread 'sys::in_addr_convertion' panicked at 'assertion failed: `(left == right)`
  left: `2130706433`,
 right: `16777343`', src/sys/unix.rs:1450:5
 ```
 
The test passes on on all little-endian architectures I have access to.

- Looking at the code, the `to_in_addr` function uses `from_ne_bytes` (equivalent to big-endian?).
- According to std sources, `Ipv4Addr::octets()` returns the bytes in network-endian order, so that's good.

I think the explicitly "shift-constructed" expected u32 value has the wrong byte order for BE systems, which is apparent when looking at the u32s in binary:

```
left:   01111111 00000000 00000000 00000001
right:  00000001 00000000 00000000 01111111
```

The "right" one is the "expected" value I get when calculating `127 | 1 << 24` manually on a little-endian system, but the "left" value definitely has the reversed byte order when constructing it manually on a big-endian system.

### PR DESCRIPTION
**Pull request by decathorpe**
Previously, the "expected" values were implicitly encoded in
little-endian byte order with bitwise operators to match the
native (little-endian) byte order of u32s on most architectures.
This test obviously breaks when the byte order is big-endian.

Instead, convert the tested address u32 into "network endian"
bytes, and provide the expected value as an array of numbers,
as well. This has the added benefit that it's much clearer to
read, and the numbers in the array also match the ones that are
used in the Ipv4Addr constructor, and no longer have to be
constructed manually, in reverse order, with bitwise operators.

Fixes #216

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -1449,13 +1449,13 @@ fn in_addr_convertion() {
     let raw = to_in_addr(&ip);
     // NOTE: `in_addr` is packed on NetBSD and it's unsafe to borrow.
     let a = raw.s_addr;
-    assert_eq!(a, 127 | 1 << 24);
+    assert_eq!(a, u32::from_ne_bytes([127, 0, 0, 1]));
     assert_eq!(from_in_addr(raw), ip);
 
     let ip = Ipv4Addr::new(127, 34, 4, 12);
     let raw = to_in_addr(&ip);
     let a = raw.s_addr;
-    assert_eq!(a, 127 << 0 | 34 << 8 | 4 << 16 | 12 << 24);
+    assert_eq!(a, u32::from_ne_bytes([127, 34, 4, 12]));
     assert_eq!(from_in_addr(raw), ip);
 }
 

### CODE REVIEW

File: src/sys/unix.rs [hunk 1]
```diff
@@ -1449,13 +1449,13 @@ fn in_addr_convertion() {
     let raw = to_in_addr(&ip);
     // NOTE: `in_addr` is packed on NetBSD and it's unsafe to borrow.
     let a = raw.s_addr;
-    assert_eq!(a, 127 | 1 << 24);
+    assert_eq!(a.to_ne_bytes(), [127, 0, 0, 1]);
```
Thomasdezeeuw, line 1452: The test is about checking the value of `a`, so could you move the conversion to the expected value?

decathorpe: So, changing this line to something like `assert_eq!(a, u32::from_ne_bytes([127, 0, 0, 1]))?

Thomasdezeeuw: Yes that should work.

decathorpe: Done, I've changed both the asserts to compare against plain `a`.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @decathorpe.

Issue #212, opened 2021-03-13
Tracking issue for socket2 based standard library


Thomasdezeeuw:
I intend to use socket2 in the standard library, however the crate is current not ready for it. This issue tracks all the issues that need addressing before the standard library can be based on socket2.

* [ ] Additional maintainers: #211.
* [ ] Proper documentation of supported OS (versions): #127.
* [ ] Additional testing: #156.

If you have additional issue that need to be added to this list please open a new issue and link to this one.

Thomasdezeeuw:
I no longer have time for this.

PR #222, opened 2021-04-25
Allow for niche optimization on Unix platforms

KaiJewson:
The standard library's file descriptor types have a niche of -1, allowing `Option<T>` to have the same size as `T`. By storing a `TcpStream` internally instead of a raw file descriptor, `Socket` can also take advantage of this. Currently, this incurs a slightly cost as `{from, as, into}_raw_{fd, socket}` won't be inlined, however if [this Rust PR](https://github.com/rust-lang/rust/pull/84541) is merged it will be. Storing a `TcpStream` also allows for closing to be implemented by the standard library instead of this one, which is a nice benefit.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -68,8 +68,25 @@ use crate::{MaybeUninitSlice, RecvFlags};
 /// # Ok(()) }
 /// ```
 pub struct Socket {
-    // The `sys` module most have access to the socket.
-    pub(crate) inner: sys::Socket,
+    inner: Inner,
+}
+
+/// Store a `TcpStream` internally to take advantage of its niche optimizations on Unix platforms.
+pub(crate) type Inner = std::net::TcpStream;
+
+// The `sys` module must have access to these functions.
+impl Socket {
+    pub(crate) fn from_raw(raw: sys::Socket) -> Socket {
+        Socket {
+            inner: sys::socket_from_raw(raw),
+        }
+    }
+    pub(crate) fn as_raw(&self) -> sys::Socket {
+        sys::socket_as_raw(&self.inner)
+    }
+    pub(crate) fn into_raw(self) -> sys::Socket {
+        sys::socket_into_raw(self.inner)
+    }
 }
 
 impl Socket {
@@ -94,7 +111,7 @@ impl Socket {
     /// Windows and simply creates a new socket, no other configuration is done.
     pub fn new_raw(domain: Domain, ty: Type, protocol: Option<Protocol>) -> io::Result<Socket> {
         let protocol = protocol.map(|p| p.0).unwrap_or(0);
-        sys::socket(domain.0, ty.0, protocol).map(|inner| Socket { inner })
+        sys::socket(domain.0, ty.0, protocol).map(Socket::from_raw)
     }
 
     /// Creates a pair of sockets which are connected to each other.
@@ -135,7 +152,7 @@ impl Socket {
     ) -> io::Result<(Socket, Socket)> {
         let protocol = protocol.map(|p| p.0).unwrap_or(0);
         sys::socketpair(domain.0, ty.0, protocol)
-            .map(|fds| (Socket { inner: fds[0] }, Socket { inner: fds[1] }))
+            .map(|[a, b]| (Socket::from_raw(a), Socket::from_raw(b)))
     }
 
     /// Binds this socket to the specified address.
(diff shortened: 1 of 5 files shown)


Thomasdezeeuw:
I actually plan to use socket2 in std lib (issue #212), so this change would make it a lot harder. Do you know if there are any plans to expose the valid range of an int (i.e. what `TcpStream` is already using) as a public API?

KaiJewson:
I do not know of any plans for that, no. This niche actually makes things difficult because it would probably be considered a regression for std to remove the niche. Excluding adding more things to the language, we could hide `Socket` behind a `std` feature flag, and have libstd use `SockRef` instead so it can implement its own niche optimization. I don't know if that would cause a circular dependency though - potentially if `-Zbuild-std` is enabled?

src/socket.rs
```diff
@@ -68,8 +68,25 @@ use crate::{MaybeUninitSlice, RecvFlags};
 /// # Ok(()) }
 /// ```
 pub struct Socket {
-    // The `sys` module most have access to the socket.
-    pub(crate) inner: sys::Socket,
+    inner: Inner,
+}
+
+/// Store a `TcpStream` internally to take advantage of its niche optimizations on Unix platforms.
+pub(crate) type Inner = std::net::TcpStream;
```
Thomasdezeeuw, line 75: Does `File` has the same optimisation? Perhaps using `TcpStream` is misleading...

src/socket.rs
```diff
@@ -68,8 +68,25 @@ use crate::{MaybeUninitSlice, RecvFlags};
...
+
+// The `sys` module must have access to these functions.
+impl Socket {
+    pub(crate) fn from_raw(raw: sys::Socket) -> Socket {
```
Thomasdezeeuw, line 79: Nit: can you merge this block with the one below?

src/socket.rs
Thomasdezeeuw: Can you remove the `Inner` type and create `Socket` directly in the `sys::socket_from_raw` module directly? Or is there a specific reason you need the type?

src/sys/unix.rs
```diff
@@ -482,6 +482,16 @@ impl SockAddr {
 
 pub(crate) type Socket = c_int;
 
+pub(crate) fn socket_from_raw(socket: Socket) -> crate::socket::Inner {
+    unsafe { crate::socket::Inner::from_raw_fd(socket) }
+}
```
Thomasdezeeuw, line 487: Nit: empty lines after function bodys. Doesn't rustfmt enforce this?

src/sys/windows.rs
```diff
@@ -190,6 +190,16 @@ fn init() {
 
 pub(crate) type Socket = sock::SOCKET;
 
+pub(crate) fn socket_from_raw(socket: Socket) -> crate::socket::Inner {
+    unsafe { crate::socket::Inner::from_raw_socket(socket as RawSocket) }
+}
```
Thomasdezeeuw, line 195: Nit: empty lines.

Thomasdezeeuw (commented):
I think it's best to forward with this, we'll figure the use-in-std stuff later.

I've reviewed the changes, they look good. Two points and some small things (prefixed with nit). I do have one question: does `self.as_raw` generate the same code as `self.inner` in all functions, i.e. is `as_raw` always inlined?

src/socket.rs
KaiJewson: Windows can't use a `File` because that stores a handle instead of a socket. And I think it's better to use `TcpStream` on both Windows and Unix, because it's possible that in the future Windows will get this niche optimization as well (since -1 is not a valid socket value I don't think).

The reason I used `Inner` was so that `TcpStream` could only be stated once in the code, where we can explain what it does. If someone sees:

```rust
pub(crate) fn socket_from_raw(socket: Socket) -> crate::Socket {
    crate::Socket { inner: unsafe { TcpStream::from_raw_fd(socket) } }
}
```

that's slightly more confusing since there's a random unexplained `TcpStream` being created in a place quite far from `Socket`'s declaration and the corresponding explanation. But I don't really mind, I'm happy to change it.

src/sys/unix.rs
KaiJewson: @Thomasdezeeuw There is an unstable Rustfmt option [`blank_lines_lower_bound`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#blank_lines_lower_bound) which enforces this, you might be interested in enabling it.

src/socket.rs
Thomasdezeeuw: 👍 clear.

Thomasdezeeuw (approved):

src/sys/unix.rs
Thomasdezeeuw: Doesn't seem to work right just quite yet: https://github.com/rust-lang/rustfmt/issues/3382#issuecomment-825168920, perhaps once it becomes stable.

Thomasdezeeuw:
FreeBSD failure seems unrelated, so I'm merging.

Thomasdezeeuw:
Thanks @KaiJewson.

PR #224, opened 2021-04-28
Add support for type-of-service option

aym-v:
Some network applications depend on proper marking of outgoing packets with ToS/DSCP fields. Routing protocols, management traffic, VoIP applications, etc., all depend on the ToS field.

ToS can be set at the socket level using the setsockopt call, using the IPPROTO_IP level and the IP_TOS option. There is already `ttl` and `set_ttl `methods to get and set the TTL field. 

This PR adds a `tos` and `set_tos `methods.


Thomasdezeeuw (approved):
LGTM.

The first sentence of the Windows documentation does say "Do not use." (https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options, search for IP_TOS). Maybe we should echo that in the docs?

aym-v:
Indeed, I added a disclaimer linking to the Windows doc. 
Related: https://github.com/golang/go/issues/42728

Thomasdezeeuw (approved):

aym-v:
Since not all target OSes support `IP_TOS`, I will add a commit to conditionally enable the option

Thomasdezeeuw (approved):

Thomasdezeeuw:
@aym-v do you have time to update this? I could take over if not.

aym-v:
@Thomasdezeeuw I was not able to reproduce the CI failures on my local computer by using the same commands. That's weird

Thomasdezeeuw:
@aym-v did you rebase on master? `git fetch origin && git rebase origin/master`, assuming `origin` points to this repo.

aym-v:
@Thomasdezeeuw Indeed, I did not do it. I just rebased the branch on master and updated the failing parts.

Thomasdezeeuw:
Fuchsia doesn't seem to have `libc::IP_TOS` defined. It could be that Fuchsia doesn't define/support the constant, or that the `libc` crate simply doesn't have it (yet). If you want, you can add it to the `libc` crate, but I would be ok with dropping Fuchsia support from this change as well, then we can open an issue for it and do it later, up to you @aym-v.

aym-v:
Fortunately, fuschia supports these constants, I just opened a PR on libc to expose them!

Thomasdezeeuw:
Thanks @aym-v! I didn't release the libc pr was merged some time ago.

PR #227, opened 2021-04-29
support setting IP_TRANSPARENT on sockets

pepinns:
This allows you to set IP_TRANSPARENT on sockets for linux.

To setup for a local proxy you can use the following:
ref: https://powerdns.org/tproxydoc/tproxy.md.html

```
    export OUTBOUND_PORT=3100
    export PROXY_PORT=3000
    # move anything marked with a 1 --> table #100
    sudo ip rule add fwmark 1 lookup 100
    # table 100, treat all addrs as local loopback
    sudo ip route add local 0.0.0.0/0 dev lo table 100

    # all inbound traffic on $OUTBOUND_PORT should forward to localhost $PROXY_PORT transparently
    sudo iptables -t mangle -I POSTROUTING -p tcp --dport $OUTBOUND_PORT -j TPROXY --tproxy-mark 0x1/0x1 --on-port=$PROXY_PORT --on-ip=127.0.0.1

    # all output traffic for $OUTBOUND_PORT will get marked and sent to table #100 (loopback)
    sudo iptables -t mangle -A OUTPUT -p tcp -m tcp --dport $OUTBOUND_PORT -j MARK --set-mark 1
```

Once done, then newly accept()'ed sockets will have the local_addr() set to the originally requested address and not the currently bound listening address.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -884,41 +884,6 @@ impl Socket {
         }
     }
 
-    /// IP_TRANSPARENT (since Linux 2.6.24)
-    /// Setting this boolean option enables transparent proxying
-    /// on this socket.  This socket option allows the calling
-    /// application to bind to a nonlocal IP address and operate
-    /// both as a client and a server with the foreign address as
-    /// the local endpoint.  NOTE: this requires that routing be
-    /// set up in a way that packets going to the foreign address
-    /// are routed through the TProxy box (i.e., the system
-    /// hosting the application that employs the IP_TRANSPARENT
-    /// socket option).  Enabling this socket option requires
-    /// superuser privileges (the CAP_NET_ADMIN capability).
-    ///
-    /// TProxy redirection with the iptables TPROXY target also
-    /// requires that this option be set on the redirected socket.
-    /// this feature is only available on linux
-    #[cfg(any(target_os = "linux"))]
-    pub fn set_ip_transparent(&self, transparent: bool) -> io::Result<()> {
-        unsafe {
-            setsockopt(
-                self.inner,
-                sys::IPPROTO_IP,
-                libc::IP_TRANSPARENT,
-                transparent as c_int,
-            )
-        }
-    }
-    /// Get whether the IP_TRANSPARENT is set for this socket.
-    #[cfg(any(target_os = "linux"))]
-    pub fn ip_transparent(&self) -> io::Result<bool> {
-        unsafe {
-            getsockopt::<c_int>(self.inner, sys::IPPROTO_IP, libc::IP_TRANSPARENT)
-                .map(|transparent| transparent != 0)
-        }
-    }
-
     /// Get the value of the `SO_SNDBUF` option on this socket.
     ///
     /// For more information about this option, see [`set_send_buffer_size`].
@@ -990,6 +955,46 @@ fn into_linger(duration: Option<Duration>) -> sys::linger {
 /// * Linux: <https://man7.org/linux/man-pages/man7/ip.7.html>
 /// * Windows: <https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options>
 impl Socket {
+    /// Get the value of the `IP_TRANSPARENTF` option on this socket.
+    ///
+    /// For more information about this option, see [`set_ip_transparent`].
+    ///
+    /// [`ip_transparent`]: Socket::set_ip_transparent
+    #[cfg(all(feature = "all", target_os = "linux"))]
+    pub fn ip_transparent(&self) -> io::Result<bool> {
+        unsafe {
+            getsockopt::<c_int>(self.inner, sys::IPPROTO_IP, libc::IP_TRANSPARENT)
+                .map(|transparent| transparent != 0)
+        }
+    }
+
+    /// IP_TRANSPARENT (since Linux 2.6.24)
+    /// Setting this boolean option enables transparent proxying
+    /// on this socket.  This socket option allows the calling
+    /// application to bind to a nonlocal IP address and operate
+    /// both as a client and a server with the foreign address as
+    /// the local endpoint.  NOTE: this requires that routing be
+    /// set up in a way that packets going to the foreign address
+    /// are routed through the TProxy box (i.e., the system
+    /// hosting the application that employs the IP_TRANSPARENT
+    /// socket option).  Enabling this socket option requires
+    /// superuser privileges (the CAP_NET_ADMIN capability).
+    ///
+    /// TProxy redirection with the iptables TPROXY target also
+    /// requires that this option be set on the redirected socket.
+    /// this feature is only available on linux
+    #[cfg(all(feature = "all", target_os = "linux"))]
+    pub fn set_ip_transparent(&self, transparent: bool) -> io::Result<()> {
+        unsafe {
+            setsockopt(
+                self.inner,
+                sys::IPPROTO_IP,
+                libc::IP_TRANSPARENT,
+                transparent as c_int,
+            )
+        }
+    }
+
     /// Join a multicast group using `IP_ADD_MEMBERSHIP` option on this socket.
     ///
     /// This function specifies a new multicast group for this socket to join.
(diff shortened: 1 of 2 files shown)


src/socket.rs
```diff
@@ -861,6 +861,41 @@ impl Socket {
         }
     }
 
+    /// IP_TRANSPARENT (since Linux 2.6.24)
```
Thomasdezeeuw, line 864: Could you following the convention of the other settings. I.e. Set value for the `IP_TRANSPARENT` option on this socket.

src/socket.rs
```diff
@@ -861,6 +861,41 @@ impl Socket {
...
+        unsafe {
+            setsockopt(
+                self.inner,
+                sys::IPPROTO_IP,
+                libc::IP_TRANSPARENT,
+                transparent as c_int,
+            )
+        }
+    }
+    /// Get whether the IP_TRANSPARENT is set for this socket.
```
Thomasdezeeuw, line 890: Could you following the convention of the other settings. I.e. Get the value of the `IP_TRANSPARENT` option on this socket.

src/socket.rs
```diff
@@ -861,6 +861,41 @@ impl Socket {
...
+    #[cfg(any(target_os = "linux"))]
+    pub fn ip_transparent(&self) -> io::Result<bool> {
```
Thomasdezeeuw, line 892: Can you move this getter before the setter (`set_ip_transparent`).

src/socket.rs
```diff
@@ -861,6 +861,41 @@ impl Socket {
...
+    /// are routed through the TProxy box (i.e., the system
+    /// hosting the application that employs the IP_TRANSPARENT
+    /// socket option).  Enabling this socket option requires
+    /// superuser privileges (the CAP_NET_ADMIN capability).
+    ///
+    /// TProxy redirection with the iptables TPROXY target also
+    /// requires that this option be set on the redirected socket.
+    /// this feature is only available on linux
+    #[cfg(any(target_os = "linux"))]
+    pub fn set_ip_transparent(&self, transparent: bool) -> io::Result<()> {
```
Thomasdezeeuw, line 880: Can you move both functions to the `IPPROTO_IP` impl block before `ttl` (its sorted alphabetically): https://github.com/rust-lang/socket2/blob/eaa5f8f66233df701983ec073f1c5683019e442e/src/socket.rs#L1074-L1076

src/socket.rs
```diff
@@ -861,6 +861,41 @@ impl Socket {
...
+    /// set up in a way that packets going to the foreign address
+    /// are routed through the TProxy box (i.e., the system
+    /// hosting the application that employs the IP_TRANSPARENT
+    /// socket option).  Enabling this socket option requires
+    /// superuser privileges (the CAP_NET_ADMIN capability).
+    ///
+    /// TProxy redirection with the iptables TPROXY target also
+    /// requires that this option be set on the redirected socket.
+    /// this feature is only available on linux
+    #[cfg(any(target_os = "linux"))]
```
Thomasdezeeuw, line 879: Since this is not available on all OSs it needs to be behind the `all` feature. It requires `cfg` attribute, see https://github.com/rust-lang/socket2/blob/eaa5f8f66233df701983ec073f1c5683019e442e/src/socket.rs#L1333-L1345

tests/socket.rs
```diff
@@ -1040,6 +1040,12 @@ test!(
     mss,
     set_mss(256)
 );
+#[cfg(any(target_os = "linux"))]
```
Thomasdezeeuw, line 1043: This would also require the `all` feature `cfg`.

Thomasdezeeuw (commented):
I'm Ok with this change, but it needs to be behind the `all` feature (as it's not available on all OSs) and it needs to be moved and updated with some style changes. I've left some comments, but it shouldn't be too much work.

src/socket.rs
pepinns: Missed this one with last commit.  Do you want me to remove the text that's there and simply replace it with what you suggested?  Or add that as a header?

src/socket.rs
pepinns: Yes, moved to proper impl block in alphabetical order.

src/socket.rs
```diff
@@ -925,6 +925,48 @@ fn into_linger(duration: Option<Duration>) -> sys::linger {
 /// * Linux: <https://man7.org/linux/man-pages/man7/ip.7.html>
 /// * Windows: <https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options>
 impl Socket {
+    /// Get the value of the `IP_TRANSPARENTF` option on this socket.
```
Thomasdezeeuw, line 928: Nit: IP_TRANSPARENT**F** -> IP_TRANSPARENT.

src/socket.rs
```diff
@@ -925,6 +925,48 @@ fn into_linger(duration: Option<Duration>) -> sys::linger {
...
+    pub fn ip_transparent(&self) -> io::Result<bool> {
+        unsafe {
+            getsockopt::<c_int>(self.inner, sys::IPPROTO_IP, libc::IP_TRANSPARENT)
+                .map(|transparent| transparent != 0)
+        }
+    }
+
+    /// Set the value of the `IP_TRANSPARENTF` option on this socket.
+    ///
+    /// IP_TRANSPARENT (since Linux 2.6.24)
```
Thomasdezeeuw, line 943: Nit: you can remove this line.

Thomasdezeeuw (approved):
There a typo in the docs: IP_TRANSPARENT**F**, otherwise LGTM.

src/socket.rs
pepinns: 943 or 942?

Thomasdezeeuw: 943: "IP_TRANSPARENT (since Linux 2.6.24)", already has it above it.

src/socket.rs
```diff
@@ -925,6 +925,48 @@ fn into_linger(duration: Option<Duration>) -> sys::linger {
...
 impl Socket {
+    /// Get the value of the `IP_TRANSPARENTF` option on this socket.
+    ///
+    /// For more information about this option, see [`set_ip_transparent`].
+    ///
+    /// [`ip_transparent`]: Socket::set_ip_transparent
+    #[cfg(all(feature = "all", target_os = "linux"))]
+    pub fn ip_transparent(&self) -> io::Result<bool> {
+        unsafe {
+            getsockopt::<c_int>(self.inner, sys::IPPROTO_IP, libc::IP_TRANSPARENT)
```
Thomasdezeeuw, line 936: From the CI: can't use `self.inner` anymore, you need to use `self.as_raw()`.

pepinns: Forgive me, happy to change it, but this confuses me and I'd like to understand what's going on.

I don't see `as_raw` in the source code for this project, is that from somewhere else?  What does it do and why would my method need to be different than the others in the file?  They appear to be using the same function(s).

Changing self.inner --> self.as_raw() still compiles, but I don't really understand why.
It looks like `self.inner` is a `Socket` which is actually a `c_int` which is a `i32`.     
It also only seems to compile when used in this method, but put into other methods causes a ```no method named `as_raw` found for reference `&socket::Socket` in the current scope``` error.

Is there some implementation of `as_raw` for `Socket` or `c_int` somewhere?  
my rust-analyzer is choking on this file so thats not really helping me either.

pepinns: Does it have to do with the feature directive somehow?

Thomasdezeeuw: This is due to https://github.com/rust-lang/socket2/pull/222 which I merged yesterday, e.g. see https://github.com/rust-lang/socket2/blob/eaa5f8f66233df701983ec073f1c5683019e442e/src/socket.rs#L694-L697

It's the same as the old `.inner`, but we now use a `TcpStream` to take advantage of the niche representation of that type.

pepinns: Ok, rebased the branch against master and things look OK now.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @pepinns!

PR #228, opened 2021-05-01
Support infallible SockAddr creation

KaiJewson:
- I renamed `SockAddr::init` to `SockAddr::try_init`, to support the addition of an infallible `SockAddr::init`.
- `SockAddr::vsock` can now be made to be infallible by using `SockAddr::init`.
- `SockAddr::try_init` now supports a generic error type, not just `io::Error` for additional flexibility.

### DIFF
--- a/src/sockaddr.rs
+++ b/src/sockaddr.rs
@@ -32,33 +32,42 @@ impl SockAddr {
     ///
     /// # Examples
     ///
-    #[cfg_attr(target_os = "linux", doc = "```")]
-    #[cfg_attr(not(target_os = "linux"), doc = "```ignore")]
+    /// ```
+    /// # fn main() -> std::io::Result<()> {
+    /// # #[cfg(unix)] {
     /// use std::io;
     /// use std::mem;
+    /// use std::os::unix::io::AsRawFd;
     ///
-    /// use socket2::{SockAddr, Socket, Domain, Type, Protocol};
+    /// use socket2::{SockAddr, Socket, Domain, Type};
     ///
-    /// # fn main() -> io::Result<()> {
-    /// // Initialise a netlink socket.
-    /// let mut storage: libc::sockaddr_storage = unsafe { mem::zeroed() };
-    /// let address: &mut libc::sockaddr_nl = unsafe { &mut *<*mut _>::cast(&mut storage) };
-    /// address.nl_family = libc::AF_NETLINK as libc::sa_family_t;
-    /// let address = unsafe { SockAddr::new(storage, mem::size_of::<libc::sockaddr_nl>() as _) };
+    /// let socket = Socket::new(Domain::IPV4, Type::STREAM, None)?;
     ///
-    /// let socket = Socket::new(
-    ///     Domain::from(libc::AF_NETLINK),
-    ///     Type::from(libc::SOCK_DGRAM),
-    ///     Some(Protocol::from(libc::NETLINK_GENERIC)),
-    /// )?;
-    /// socket.bind(&address)?;
-    /// # drop(socket);
+    /// // Initialise a `SocketAddr` byte calling `getsockname(2)`.
+    /// let mut addr_storage: libc::sockaddr_storage = unsafe { mem::zeroed() };
+    /// let mut len = 0;
+    ///
+    /// // The `getsockname(2)` system call will intiliase `storage` for
+    /// // us, setting `len` to the correct length.
+    /// let res = unsafe {
+    ///     libc::getsockname(
+    ///         socket.as_raw_fd(),
+    ///         (&mut addr_storage as *mut libc::sockaddr_storage).cast(),
+    ///         &mut len,
+    ///     )
+    /// };
+    /// if res == -1 {
+    ///     return Err(io::Error::last_os_error());
+    /// }
+    ///
+    /// let address = unsafe { SockAddr::new(addr_storage, len) };
+    /// # drop(address);
+    /// # }
     /// # Ok(())
     /// # }
-    #[doc = "```"]
-    #[must_use]
-    pub const unsafe fn new(storage: sockaddr_storage, len: socklen_t) -> Self {
-        Self { storage, len }
+    /// ```
+    pub const unsafe fn new(storage: sockaddr_storage, len: socklen_t) -> SockAddr {
+        SockAddr { storage, len }
     }
 
     /// Initialise a `SockAddr` by calling the function `init`.
@@ -77,14 +86,14 @@ impl SockAddr {
     ///
     /// # Examples
     ///
-    #[cfg_attr(unix, doc = "```")]
-    #[cfg_attr(not(unix), doc = "```ignore")]
+    /// ```
+    /// # fn main() -> std::io::Result<()> {
+    /// # #[cfg(unix)] {
     /// use std::io;
     /// use std::os::unix::io::AsRawFd;
     ///
     /// use socket2::{SockAddr, Socket, Domain, Type};
     ///
-    /// # fn main() -> io::Result<()> {
     /// let socket = Socket::new(Domain::IPV4, Type::STREAM, None)?;
     ///
     /// // Initialise a `SocketAddr` byte calling `getsockname(2)`.
@@ -100,9 +109,10 @@ impl SockAddr {
     ///     })
     /// }?;
     /// # drop(address);
+    /// # }
     /// # Ok(())
     /// # }
-    #[doc = "```"]
+    /// ```
     pub unsafe fn init<F, T>(init: F) -> io::Result<(T, SockAddr)>
     where
         F: FnOnce(*mut sockaddr_storage, *mut socklen_t) -> io::Result<T>,
--- a/src/sockaddr.rs
+++ b/src/sockaddr.rs
@@ -18,6 +18,7 @@ pub struct SockAddr {
     len: socklen_t,
 }
 
+#[allow(clippy::len_without_is_empty)]
 impl SockAddr {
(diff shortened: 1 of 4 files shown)


Thomasdezeeuw:
At the moment I don't really want to release a breaking change as v0.4 is only 2 months old, but we keep a list for possible breaking changes/improvements. So let's revert the `init` -> `try_init` name change.

Maybe we should add more traditional `new` function:
```rust
const fn new(storage: sockaddr_storage, len: socklen_t) -> SockAddr;
```

What do you think @KaiJewson?

KaiJewson:
That might be a good idea - would it suffer from the same copying problem `from_raw_parts` had?

Thomasdezeeuw:
> That might be a good idea - would it suffer from the same copying problem `from_raw_parts` had?

The original `from_raw_parts` took pointer to the address storage and explicitly copied it. The function signature I proposed takes it by value, if the function gets inlined (might require a `#[inline]` attribute) then it wouldn't copy, otherwise I think it would be implicitly copied (because `sockaddr_storage` is `Copy`).

KaiJewson:
That's great then, I will switch this PR to add that. Should it be `unsafe fn` or not?

Another thing I was considering is adding `zeroed` + `as_ptr_mut`/`storage_mut` + `len_mut`/`set_len` so that it can be constructed in-place. Or alternatively, a `fn from_inner<T>(inner: T) -> SockAddr` that fills in the `SockAddr` based on the bytes and size of `inner`.

Thomasdezeeuw:
> That's great then, I will switch this PR to add that. Should it be `unsafe fn` or not?

Good question. I think it should be because we depend on the length being correct. And the family must be set correctly as we depend on that in `as_socket`. Two good things to add to the docs as well.

> Another thing I was considering is adding `zeroed` + `as_ptr_mut`/`storage_mut` + `len_mut`/`set_len` so that it can be constructed in-place. Or alternatively, a `fn from_inner<T>(inner: T) -> SockAddr` that fills in the `SockAddr` based on the bytes and size of `inner`.

I think `new` would fill the spot nicely. Could you give an example of where `new`/`init` aren't sufficient?

KaiJewson:
> I think `new` would fill the spot nicely. Could you give an example of where `new`/`init` aren't sufficient?

`new`/`init` are always sufficient, but it's just about having a slightly nicer API. Here is some code using `new`:

```rust
let mut storage: sockaddr_storage = unsafe { mem::zeroed() };
(&mut storage as *mut sockaddr_storage).cast::<sockaddr_in>().write(sockaddr_in {
    sin_family: AF_INET,
    sin_port: 80,
    sin_addr: in_addr { s_addr: 0 },
    sin_zero: [0; 8],
});
let addr = SockAddr::new(storage, mem::size_of::<sockaddr_in>() as socklen_t);
```

And here's the same code with `from_inner`:

```rust
let addr = SockAddr::from_inner(sockaddr_in {
    sin_family: AF_INET,
    sin_port: 80,
    sin_addr: in_addr { s_addr: 0 },
    sin_zero: [0; 8],
});
```

src/sockaddr.rs
```diff
@@ -12,24 +12,68 @@ use winapi::shared::ws2ipdef::SOCKADDR_IN6_LH_u;
...
+    ///     Domain::from(libc::AF_NETLINK),
+    ///     Type::from(libc::SOCK_DGRAM),
+    ///     Some(Protocol::from(libc::NETLINK_GENERIC)),
+    /// )?;
+    /// socket.bind(&address)?;
+    /// # drop(socket);
+    /// # Ok(())
+    /// # }
+    #[doc = "```"]
+    #[must_use]
```
Thomasdezeeuw, line 59: This isn't required.

src/sockaddr.rs
```diff
@@ -12,24 +12,68 @@ use winapi::shared::ws2ipdef::SOCKADDR_IN6_LH_u;
...
+    /// let socket = Socket::new(
+    ///     Domain::from(libc::AF_NETLINK),
+    ///     Type::from(libc::SOCK_DGRAM),
+    ///     Some(Protocol::from(libc::NETLINK_GENERIC)),
+    /// )?;
+    /// socket.bind(&address)?;
+    /// # drop(socket);
+    /// # Ok(())
+    /// # }
+    #[doc = "```"]
```
Thomasdezeeuw, line 58: You can just use:

    /// ```

src/sys/unix.rs
```diff
@@ -444,6 +444,11 @@ impl SockAddr {
     /// Constructs a `SockAddr` with the family `AF_VSOCK` and the provided CID/port.
     ///
     /// This function is only available on Linux.
+    ///
+    /// # Errors
+    ///
+    /// This function can never fail. In a future version of this library it will be made
+    /// infallible.
```
Thomasdezeeuw, line 451: This shouldn't be part of this pr, let's try to keep it small.

src/socket.rs
```diff
@@ -655,6 +655,7 @@ fn set_common_type(ty: Type) -> Type {
 
 /// Set `FD_CLOEXEC` and `NOSIGPIPE` on the `socket` for platforms that need it.
 #[inline(always)]
+#[allow(clippy::unnecessary_wraps)]
```
Thomasdezeeuw, line 658: This doesn't seem necessary.

src/sockaddr.rs
```diff
@@ -58,7 +102,7 @@ impl SockAddr {
     /// # drop(address);
     /// # Ok(())
     /// # }
-    /// ```
+    #[doc = "```"]
```
Thomasdezeeuw, line 105: Please revert this.

src/sockaddr.rs
```diff
@@ -12,24 +12,68 @@ use winapi::shared::ws2ipdef::SOCKADDR_IN6_LH_u;
...
+    ///
+    /// # Safety
+    ///
+    /// - The `SockAddr`'s address family must match its contents.
+    /// - The given length of the `SockAddr` must be correct.
+    ///
+    /// # Examples
+    ///
+    #[cfg_attr(target_os = "linux", doc = "```")]
+    #[cfg_attr(not(target_os = "linux"), doc = "```ignore")]
```
Thomasdezeeuw, line 36: Can you change the example to something that works on all OSs? You can use the same one from `Socket::init`.

src/sockaddr.rs
```diff
@@ -12,24 +12,68 @@ use winapi::shared::ws2ipdef::SOCKADDR_IN6_LH_u;
...
+    ///     Type::from(libc::SOCK_DGRAM),
+    ///     Some(Protocol::from(libc::NETLINK_GENERIC)),
+    /// )?;
+    /// socket.bind(&address)?;
+    /// # drop(socket);
+    /// # Ok(())
+    /// # }
+    #[doc = "```"]
+    #[must_use]
+    pub const unsafe fn new(storage: sockaddr_storage, len: socklen_t) -> Self {
```
Thomasdezeeuw, line 60: Nit: Can you use `SockAddr` instead of `Self`, I always use `Self` for builder patterns, which this isn't.

src/sockaddr.rs
KaiJewson: I wrote it this way so that syntax highlighting doesn't break in my editor; with the old method, all doc comments below that would be treated as Rust code. Although I just realized that the example itself could have the appropriate `#![cfg(...)]` at the top and use regular triple-backtick on all platforms, so I'll switch to that instead.

src/socket.rs
KaiJewson: It prevents that Clippy warning from firing on platforms where the function is a no-op. Should I remove it?

Thomasdezeeuw: That's why it didn't trigger on macOS. In that case you can leave it.

src/sockaddr.rs
```diff
@@ -12,35 +12,88 @@ use winapi::shared::ws2ipdef::SOCKADDR_IN6_LH_u;
...
+    /// - The address family of the socket address must match its contents.
+    /// - The length of the socket address must be correct.
     ///
     /// # Examples
     ///
-    #[cfg_attr(unix, doc = "```")]
-    #[cfg_attr(not(unix), doc = "```ignore")]
+    /// ```
+    /// # fn main() -> std::io::Result<()> {
+    /// # #[cfg(unix)] {
```
Thomasdezeeuw, line 91: That's a neat solution.

src/sockaddr.rs
```diff
@@ -12,35 +12,88 @@ use winapi::shared::ws2ipdef::SOCKADDR_IN6_LH_u;
...
+    /// use std::mem;
+    /// use std::os::unix::io::AsRawFd;
+    ///
+    /// use socket2::{SockAddr, Socket, Domain, Type};
+    ///
+    /// let socket = Socket::new(Domain::IPV4, Type::STREAM, None)?;
+    ///
+    /// // Initialise a `SocketAddr` byte calling `getsockname(2)`.
+    /// let mut addr_storage: libc::sockaddr_storage = unsafe { mem::zeroed() };
+    /// let mut len = 0;
```
Thomasdezeeuw, line 48: I think this should be `size_of_val(&addr_storage)` as otherwise the OS think it can only write zero bytes, and thus doesn't write the address.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Merged as c24bb6820030c4260876be5ab501e98ffb3058fd, 0624146954f609ec03e94202379b8da447c8b669 and 99f508958eb3d2533b38040eaaaa9284a6a84951. Thanks @KaiJewson.

PR #230, opened 2021-05-03
Document `SockAddr::vsock` as infallible

KaiJewson:
Split off from #228. `SockAddr::vsock` can never fail, so this makes that clear to users while avoiding breaking changes.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -444,6 +444,11 @@ impl SockAddr {
     /// Constructs a `SockAddr` with the family `AF_VSOCK` and the provided CID/port.
     ///
     /// This function is only available on Linux.
+    ///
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
@elmarco you originally wrote this code, do you know why `SockAddr::vsock` returned an I/O result in the first place?

elmarco:
> @elmarco you originally wrote this code, do you know why `SockAddr::vsock` returned an I/O result in the first place?

I don't remember. Probably I adapted it based on other socket types returning Result, for consistency.

I suppose this may be caught by clippy, and I don't see much problem in making it infaillible.

thanks

Thomasdezeeuw (approved):

PR #232, opened 2021-05-03
Set noinherit on accepted sockets on Windows

KaiJewson:
Sockets created by `accept` will be inheritable by default, and this library does not make them non-inheritable even when calling `Socket::accept`. This change fixes that.

### DIFF
--- a/tests/socket.rs
+++ b/tests/socket.rs
@@ -154,19 +154,19 @@ fn set_nonblocking() {
     assert_nonblocking(&socket, false);
 }
 
-fn assert_default_flags(socket: &Socket, default_flags: bool) {
+fn assert_common_flags(socket: &Socket, expected: bool) {
     #[cfg(unix)]
-    assert_close_on_exec(socket, default_flags);
+    assert_close_on_exec(socket, expected);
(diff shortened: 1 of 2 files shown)


tests/socket.rs
```diff
@@ -154,26 +154,43 @@ fn set_nonblocking() {
     assert_nonblocking(&socket, false);
 }
 
-#[test]
-fn default_flags() {
-    let socket = Socket::new(Domain::IPV4, Type::STREAM, None).unwrap();
+fn assert_default_flags(socket: &Socket, default_flags: bool) {
```
Thomasdezeeuw, line 157: Nit: can you change this to `assert_common_flags` to match `set_common_flags` naming, maybe even add a reference to it as comment.

Thomasdezeeuw: Also change `default_flags` to `expected`, since `default` wouldn't make sense anymore.

Thomasdezeeuw (approved):
LGTM, one small nit.

Thomasdezeeuw:
Thanks @KaiJewson.

PR #233, opened 2021-05-03
Use doc_cfg

### LINKED ISSUE #167: Use doc_cfg on docs.rs
Doc: https://doc.rust-lang.org/unstable-book/language-features/doc-cfg.html

An example from Mio: https://docs.rs/mio/0.7.6/mio/, https://github.com/tokio-rs/mio/blob/ec28a8dcdd09bf2d7ad4030b7330c86834827124/src/lib.rs#L72.

KaiJewson:
I tried to implement it so that as many functions are displayed on the documentation as possible regardless of the platform the docs themselves were built on. This is not possible in all cases, for example the `sys/*.rs` modules had to be left out on other platforms, but at least the union of Linux and Windows docs covers everything.

Fixes #167.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -197,10 +197,12 @@ impl Type {
 
     /// Type corresponding to `SOCK_SEQPACKET`.
     #[cfg(feature = "all")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "all")))]
     pub const SEQPACKET: Type = Type(sys::SOCK_SEQPACKET);
 
     /// Type corresponding to `SOCK_RAW`.
     #[cfg(all(feature = "all", not(target_os = "redox")))]
+    #[cfg_attr(docsrs, doc(cfg(all(feature = "all", not(target_os = "redox")))))]
     pub const RAW: Type = Type(sys::SOCK_RAW);
 }
 
@@ -256,6 +258,7 @@ impl From<Protocol> for c_int {
 ///
 /// Flags provide additional information about incoming messages.
 #[cfg(not(target_os = "redox"))]
+#[cfg_attr(docsrs, doc(cfg(not(target_os = "redox"))))]
 #[derive(Copy, Clone, Eq, PartialEq)]
 pub struct RecvFlags(c_int);
 
@@ -369,6 +372,20 @@ impl TcpKeepalive {
             windows,
         )
     ))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(
+            feature = "all",
+            any(
+                target_os = "freebsd",
+                target_os = "fuchsia",
+                target_os = "linux",
+                target_os = "netbsd",
+                target_vendor = "apple",
+                windows,
+            )
+        )))
+    )]
     pub const fn with_interval(self, interval: Duration) -> Self {
         Self {
             interval: Some(interval),
@@ -383,6 +400,7 @@ impl TcpKeepalive {
     #[cfg(all(
         feature = "all",
         any(
+            doc,
             target_os = "freebsd",
             target_os = "fuchsia",
             target_os = "linux",
@@ -390,6 +408,19 @@ impl TcpKeepalive {
             target_vendor = "apple",
         )
     ))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(
+            feature = "all",
+            any(
+                target_os = "freebsd",
+                target_os = "fuchsia",
+                target_os = "linux",
+                target_os = "netbsd",
+                target_vendor = "apple",
+            )
+        )))
+    )]
     pub const fn with_retries(self, retries: u32) -> Self {
         Self {
             retries: Some(retries),
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -963,7 +963,7 @@ impl Socket {
     ///
     /// For more information about this option, see [`set_ip_transparent`].
     ///
-    /// [`ip_transparent`]: Socket::set_ip_transparent
+    /// [`set_ip_transparent`]: Socket::set_ip_transparent
     #[cfg(any(doc, all(feature = "all", target_os = "linux")))]
     #[cfg_attr(docsrs, doc(cfg(all(feature = "all", target_os = "linux"))))]
     pub fn ip_transparent(&self) -> io::Result<bool> {
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -118,11 +118,8 @@ impl Socket {
     ///
     /// This function sets the same flags as in done for [`Socket::new`],
     /// [`Socket::pair_raw`] can be used if you don't want to set those flags.
-    ///
-    /// # Notes
-    ///
-    /// This function is only available on Unix.
-    #[cfg(all(feature = "all", unix))]
+    #[cfg(any(doc, all(feature = "all", unix)))]
+    #[cfg_attr(docsrs, doc(cfg(all(feature = "all", unix))))]
     pub fn pair(
         domain: Domain,
         ty: Type,
@@ -138,11 +135,8 @@ impl Socket {
     /// Creates a pair of sockets which are connected to each other.
     ///
     /// This function corresponds to `socketpair(2)`.
-    ///
-    /// # Notes
-    ///
-    /// This function is only available on Unix.
-    #[cfg(all(feature = "all", unix))]
+    #[cfg(any(doc, all(feature = "all", unix)))]
+    #[cfg_attr(docsrs, doc(cfg(all(feature = "all", unix))))]
     pub fn pair_raw(
         domain: Domain,
         ty: Type,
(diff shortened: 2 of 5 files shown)


src/lib.rs
```diff
@@ -383,13 +400,27 @@ impl TcpKeepalive {
     #[cfg(all(
         feature = "all",
         any(
+            doc,
```
Thomasdezeeuw, line 403: Is this supposed to be here?

src/socket.rs
```diff
@@ -118,11 +118,8 @@ impl Socket {
     ///
     /// This function sets the same flags as in done for [`Socket::new`],
     /// [`Socket::pair_raw`] can be used if you don't want to set those flags.
-    ///
-    /// # Notes
-    ///
-    /// This function is only available on Unix.
-    #[cfg(all(feature = "all", unix))]
+    #[cfg(any(doc, all(feature = "all", unix)))]
```
Thomasdezeeuw, line 121: I don't think the `any(doc, ...` part works., same for for a number of other functions.

src/sockref.rs
```diff
@@ -20,8 +20,9 @@ use crate::Socket;
 /// the caller must ensure the file descriptor/socket is a valid.
 ///
 /// [`TcpStream`]: std::net::TcpStream
-/// [`AsRawFd`]: std::os::unix::io::AsRawFd
-/// [`AsRawSocket`]: std::os::windows::io::AsRawSocket
+// Don't use intra-doc links because they won't build on every platform.
```
Thomasdezeeuw, line 23: What do you mean by this? It should like to docs on rust-lang.org anyway.

src/sys/unix.rs
```diff
@@ -207,70 +211,56 @@ impl_debug!(
...
+    )]
     libc::AF_PACKET,
     #[cfg(any(target_os = "android", target_os = "linux"))]
+    #[cfg_attr(docsrs, doc(cfg(any(target_os = "android", target_os = "linux"))))]
     libc::AF_VSOCK,
     libc::AF_UNSPEC, // = 0.
 );
 
 /// Unix only API.
+#[cfg(any(
```
Thomasdezeeuw, line 226: This can be removed as this always true (there all `unix`es). Note not the `cfg_attr(docrs, ...`, just this `cfg`.

src/sys/unix.rs
```diff
@@ -207,70 +211,56 @@ impl_debug!(
...
-            target_os = "dragonfly",
-            target_os = "freebsd",
-            target_os = "fuchsia",
-            target_os = "illumos",
-            target_os = "linux",
-            target_os = "netbsd",
-            target_os = "openbsd"
-        )
-    ))]
+    #[cfg(feature = "all")]
```
Thomasdezeeuw, line 251: Please revert to the old `cfg` as Apple (macOS, iOS) don't have `SOCK_NONBLOCK`, same for the `cfg_attr` below.

src/sys/unix.rs
```diff
@@ -207,70 +211,56 @@ impl_debug!(
...
+    #[cfg_attr(docsrs, doc(cfg(feature = "all")))]
     pub const fn cloexec(self) -> Type {
         self._cloexec()
     }
 
-    #[cfg(any(
```
Thomasdezeeuw, line 264: Please revert, not available on Apple.

src/sys/unix.rs
```diff
@@ -207,70 +211,56 @@ impl_debug!(
...
         Type(self.0 | libc::SOCK_NONBLOCK)
     }
 
     /// Set `SOCK_CLOEXEC` on the `Type`.
-    ///
-    /// # Notes
-    ///
-    /// This function is only available on Android, DragonFlyBSD, Fuchsia,
-    /// FreeBSD, Linux, NetBSD and OpenBSD.
-    #[cfg(all(
```
Thomasdezeeuw, line 247: Please revert, not available on Apple.

src/sys/unix.rs
```diff
@@ -440,16 +429,18 @@ impl SockAddr {
     }
 }
 
+#[cfg(all(feature = "all", any(target_os = "android", target_os = "linux")))]
```
Thomasdezeeuw, line 432: Same as above, this `cfg` is not needed (the `cfg_attr` is).

src/sys/unix.rs
```diff
@@ -473,7 +464,6 @@ impl SockAddr {
 
     /// Returns this address VSOCK CID/port if it is in the `AF_VSOCK` family,
     /// otherwise return `None`.
-    #[cfg(all(feature = "all", any(target_os = "android", target_os = "linux")))]
```
Thomasdezeeuw, line 476: Revert, not available on all OSs.

src/sys/unix.rs
```diff
@@ -440,16 +429,18 @@ impl SockAddr {
...
 impl SockAddr {
     /// Constructs a `SockAddr` with the family `AF_VSOCK` and the provided CID/port.
     ///
-    /// This function is only available on Linux.
-    ///
     /// # Errors
     ///
     /// This function can never fail. In a future version of this library it will be made
     /// infallible.
-    #[cfg(all(feature = "all", any(target_os = "android", target_os = "linux")))]
```
Thomasdezeeuw, line 452: Please revert, not available everywhere.

src/sys/unix.rs
```diff
@@ -999,11 +1006,8 @@ impl crate::Socket {
     }
 
     /// Sets `SO_NOSIGPIPE` on the socket.
-    ///
-    /// # Notes
-    ///
-    /// Only supported on Apple platforms (`target_vendor = "apple"`).
-    #[cfg(all(feature = "all", target_vendor = "apple"))]
+    #[cfg(all(feature = "all", any(doc, target_vendor = "apple")))]
```
Thomasdezeeuw, line 1009: Remove `doc`.

Thomasdezeeuw (commented):
Thanks for doing this. I was putting this off in the hopes it would become a default when using `cfg`, but that hasn't happened yet. Overall this look good, just some small changes.

This needs a change to `Cargo.toml` to set the `docsrs` flag, see Mio: https://github.com/tokio-rs/mio/blob/ec28a8dcdd09bf2d7ad4030b7330c86834827124/Cargo.toml#L67-L81. You can test it locally using `RUSTDOCFLAGS="--cfg docsrs" cargo doc`, that is what docs.rs would use.

You've changed/removed some `cfg` attributes on some constants/functions which are not valid, e.g. on `Type::nonblocking`, these need to be reverted as not all OSs support all function declared here.

Sometimes you include a `doc` in the `cfg` attribute, last I checked that didn't work and you need to do everything manually (as you did) with `doc(cfg(...))`.

src/lib.rs
KaiJewson: Yes, its purpose is to have this function be documented even on platforms that do not support it (to reduce the need for users of the library to view the docs on multiple platforms).

src/socket.rs
KaiJewson: Same as above - this will allow this function to appear in documentation generated on a Windows machine (rustdoc doesn't check function bodies so its OK). So someone will not have to check every platform's docs to see that this function exists.

src/sockref.rs
KaiJewson: When the docs are generated on a Unix machine, `AsRawSocket` doesn't exist so Rustdoc isn't able to find the link to it. What std does to have `AsRawSocket` and `AsRawFd` both exist even when documented on Unix (normally only `AsRawFd` would exist because that's the current platform) is very magic, and this magic isn't inherited by the documentation of external crates (so we only get access to one while documenting).

src/sys/unix.rs
KaiJewson: This won't be enabled on Apple because the above `#[cfg]` that applies to this entire impl block excludes Apple.

I just factored out the common "android, dragonfly, freebsd, fuchsia, illumos, linux, netbsd, openbsd" cfg that previously applied to every item in the impl block individually to the impl block itself, to avoid having to write out the 23-line long attributes out each time.

src/sys/unix.rs
KaiJewson: I did a similar thing to what I did above; factor out the `#[cfg]` to the `impl` block instead of each item in the `impl` block individually. The effect of the `#[cfg]` is actually the exact same, it just avoids duplicating the attributes.

KaiJewson:
> This needs a change to Cargo.toml to set the docsrs flag

I believe [you have already made this change](https://github.com/rust-lang/socket2/pull/72) :P. And I have tested it locally with `--cfg docsrs`, it seems to work.

I responded to your other comments in the review threads.

src/sys/unix.rs
Thomasdezeeuw: I know it's easier for now, but when we add a new function/constant for macOS we would have to revert this again. I know it's overly verbose but I think I prefer the `cfg` to be complete for a given function without having to consider the `cfg` on the impl blocks, I think it makes reading it easier.

Thomasdezeeuw (commented):
> > This needs a change to Cargo.toml to set the docsrs flag
> 
> I believe [you have already made this change](https://github.com/rust-lang/socket2/pull/72) :P. And I have tested it locally with `--cfg docsrs`, it seems to work.

I completely forgot I did that... :)

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @KaiJewson!

PR #234, opened 2021-05-06
Add DragonFly support for KEEPINTVL/KEEPCNT

kworr:
Those are supported on DragonFly, build fails without them.

Not sure what to do with test, mine showing:

```
---- only_v6 stdout ----
thread 'only_v6' panicked at 'assertion failed: `(left != right)`
  left: `true`,
 right: `true`: initial value and argument are the same', tests/socket.rs:1094:1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- tcp_keepalive stdout ----
thread 'tcp_keepalive' panicked at 'assertion failed: `(left == right)`
  left: `75s`,
 right: `30s`', tests/socket.rs:727:5
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -361,6 +361,7 @@ impl TcpKeepalive {
     #[cfg(all(
         feature = "all",
         any(
+            target_os = "dragonfly",
             target_os = "freebsd",
             target_os = "fuchsia",
             target_os = "linux",
@@ -383,6 +384,7 @@ impl TcpKeepalive {
     #[cfg(all(
         feature = "all",
         any(
+            target_os = "dragonfly",
             target_os = "freebsd",
             target_os = "fuchsia",
             target_os = "linux",
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -112,6 +112,7 @@ pub(crate) use libc::{
     feature = "all",
     any(
         target_os = "android",
+        target_os = "dragonfly",
         target_os = "freebsd",
         target_os = "fuchsia",
         target_os = "illumos",
--- a/tests/socket.rs
+++ b/tests/socket.rs
@@ -683,6 +683,7 @@ fn tcp_keepalive() {
     #[cfg(all(
         feature = "all",
         any(
+            target_os = "dragonfly",
             target_os = "freebsd",
             target_os = "fuchsia",
             target_os = "linux",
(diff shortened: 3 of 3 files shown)


Thomasdezeeuw (commented):
Can you also add it to `TcpKeepalive::with_interval` and `TcpKeepalive::with_retries`?

Thomasdezeeuw:
> Not sure what to do with test, mine showing:
> 
> ```
> ---- only_v6 stdout ----
> thread 'only_v6' panicked at 'assertion failed: `(left != right)`
>   left: `true`,
>  right: `true`: initial value and argument are the same', tests/socket.rs:1094:1
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> 
> ---- tcp_keepalive stdout ----
> thread 'tcp_keepalive' panicked at 'assertion failed: `(left == right)`
>   left: `75s`,
>  right: `30s`', tests/socket.rs:727:5
> ```

For the `only_v6` can you add DragonFlyBSD to the `cfg` statements here: https://github.com/rust-lang/socket2/blob/9d3e39c696e76b26af5faf0dffb7dd5b53dc0ebc/tests/socket.rs#L1093-L1097. It seems that by default DragonFly's IPv6 sockets are IPv6 only already, just like on FreeBSD and Windows. I would do it myself, but I don't have DragonFlyBSD to test it on.

I think the `tcp_keepalive` test should be fixed if add DragonFly to the `cfg` list in `TcpKeepalive::with_interval`.

kworr:
We v6 still doesn't work for me:
```
---- only_v6 stdout ----
thread 'only_v6' panicked at 'failed to set option: Os { code: 45, kind: Other, message: "Operation not supported" }', tests/socket.rs:1111:1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Thomasdezeeuw:
> We v6 still doesn't work for me:
> 
> ```
> ---- only_v6 stdout ----
> thread 'only_v6' panicked at 'failed to set option: Os { code: 45, kind: Other, message: "Operation not supported" }', tests/socket.rs:1111:1
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

That odd... does DragonFlyBSD not support setting IPv6 only?

For the sake of merging this pr you can remove DragonFlyBSD from the tests.

kworr:
Commented out DragonflyBSD in last test leaving comment that it's actually not supported.

Thomasdezeeuw:
Merged as b96fa3bc7273a6b8cd5bdae9cb27fa4fc7458b62. Thanks @kworr.

PR #241, opened 2021-06-05
Support TCP_USER_TIMEOUT socket option for various unix-likes

jmesmon:
TCP_USER_TIMEOUT allows the user to configure tcp connections to close if the data sits in buffers without getting sent for some user specified time. It's another angle at resolving stuck tcp connections.

I've enabled this for ~all~ most targets that libc::TCP_USER_TIMEOUT exists for (linux, android, ~l4re, emscripten,~ fuschia), but I've only tested it on linux

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -1597,6 +1597,57 @@ impl crate::Socket {
         ))
         .map(|_| sbytes as usize)
     }
+
+    /// Set the value of the `TCP_USER_TIMEOUT` option on this socket.
+    ///
+    /// If set, this specifies the maximum amount of time that transmitted data may remain
+    /// unacknowledged or buffered data may remain untransmitted before TCP will forcibly close the
+    /// corresponding connection.
+    #[cfg(all(feature = "all", any(target_os = "linux", target_os = "android")))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(feature = "all", any(target_os = "linux", target_os = "android"))))
+    )]
+    pub fn set_tcp_user_timeout(&self, duration: Option<Duration>) -> io::Result<()> {
+        use std::convert::TryInto;
+        fn into_millis(duration: Option<Duration>) -> io::Result<c_int> {
+            duration
+                .map_or(0, |d| d.as_millis())
+                .try_into()
+                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))
+        }
+        unsafe {
+            setsockopt(
+                self.as_raw(),
+                libc::IPPROTO_TCP,
+                libc::TCP_USER_TIMEOUT,
+                into_millis(duration)?,
+            )
+        }
+    }
+
+    /// Get the value of the `TCP_USER_TIMEOUT` option on this socket.
+    ///
+    /// For more information about this option, see [`set_tcp_user_timeout`].
+    ///
+    /// [`set_tcp_user_timeout`]: Socket::set_tcp_user_timeout
+    #[cfg(all(feature = "all", any(target_os = "linux", target_os = "android")))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(feature = "all", any(target_os = "linux", target_os = "android"))))
+    )]
+    pub fn tcp_user_timeout(&self) -> io::Result<Option<Duration>> {
+        unsafe {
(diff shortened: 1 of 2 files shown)


src/sys/unix.rs
```diff
@@ -1597,6 +1597,93 @@ impl crate::Socket {
...
+                target_os = "emscripten",
+                target_os = "fuchsia",
+                target_os = "l4re",
+                target_os = "linux",
+            )
+        )))
+    )]
+    pub fn set_tcp_user_timeout(&self, duration: Option<Duration>) -> io::Result<()> {
+        use std::convert::TryInto;
+        fn into_millis(duration: Option<Duration>) -> io::Result<c_int> {
```
Thomasdezeeuw, line 1631: Per the `TCP_USER_TIMEOUT` docs at https://man7.org/linux/man-pages/man7/tcp.7.html:

> This option takes an unsigned int as an argument.

So this should be a `libc::c_uint`.

src/sys/unix.rs
```diff
@@ -1597,6 +1597,93 @@ impl crate::Socket {
...
     }
+
+    /// Set the value of the `TCP_USER_TIMEOUT` option on this socket.
+    ///
+    /// If set, this specifies the maximum amount of time that transmitted data may remain
+    /// unacknowledged or buffered data may remain untransmitted before TCP will forcibly close the
+    /// corresponding connection.
+    #[cfg(all(
+        feature = "all",
+        any(
```
Thomasdezeeuw, line 1608: You can reduce this to Android, Fuchsia and Linux. So we have don't have any special targets for emscripten or l4re. Same for the other `cfg` attributes (also in the test).

src/sys/unix.rs
```diff
@@ -1597,6 +1597,93 @@ impl crate::Socket {
         ))
         .map(|_| sbytes as usize)
     }
+
+    /// Set the value of the `TCP_USER_TIMEOUT` option on this socket.
+    ///
+    /// If set, this specifies the maximum amount of time that transmitted data may remain
+    /// unacknowledged or buffered data may remain untransmitted before TCP will forcibly close the
+    /// corresponding connection.
```
Thomasdezeeuw, line 1605: Can you add that when `None` is used it will use the system default. That might be surprising behaviour.

src/sys/unix.rs
```diff
@@ -1597,6 +1597,93 @@ impl crate::Socket {
...
+        )))
+    )]
+    pub fn set_tcp_user_timeout(&self, duration: Option<Duration>) -> io::Result<()> {
+        use std::convert::TryInto;
+        fn into_millis(duration: Option<Duration>) -> io::Result<c_int> {
+            duration
+                .map_or(0, |d| d.as_millis())
+                .try_into()
+                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))
+        }
```
Thomasdezeeuw, line 1636: Can you remove the function and replace it with the following.

```rust
        let timeout = timeout
            .map(|to| cmp::min(to.as_millis(), libc::c_uint::MAX as u128) as libc::c_uint)
            .unwrap_or(0);
```

Thomasdezeeuw (commented):
Overall this looks good, couple small things.

src/sys/unix.rs
jmesmon: https://github.com/rust-lang/socket2/pull/241/commits/7b8473b96a2da2b15a01eb6e91aedf04ce354afd

src/sys/unix.rs
jmesmon: https://github.com/rust-lang/socket2/pull/241/commits/df14e0d8c5dd6ddc4056e0c4cad1509ea49993c8

src/sys/unix.rs
jmesmon: https://github.com/rust-lang/socket2/pull/241/commits/df14e0d8c5dd6ddc4056e0c4cad1509ea49993c8

src/sys/unix.rs
jmesmon: Was resolved by switching to clamping instead of emitting an error on oversized duration

jmesmon:
I've added 2 commits to resolve the comments.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @jmesmon.

PR #245, opened 2021-06-09
IovLen support uclibc && upgrade libc dependencies

kolapapa:


Thomasdezeeuw (approved):
I don't run uclibc myself and I don't think we test it in the CI, but I did check the libc docs and this seems fine.

Thomasdezeeuw:
Thanks @kolapapa!

PR #246, opened 2021-06-30
Add recv_from_into_buf()

### LINKED ISSUE #223: Socket::recv_from taking a &mut [u8]
Starting in version 0.4.0 there's no safe way to pass a `&mut [u8]` into recv_from, since RecvFrom takes `&mut [MaybeUninitialized<u8>]` and rustc does not understand if the library potentially writes uninitalized bytes *to* the buffer.

Would it be possible to return the signature of the function to the way it was in 0.3.0 so that we can safely call recv_from? Perhaps a recv_from_uninitialized_with_options would be something useful for hyper-tuned applications.

Currently the only way I can see to interact with recv_from when I have a &mut [u8] is to use transmute, which is very difficult to pass code review with and difficult to reason about.

agrover:
It would be nice to have a way to use `recv_from` without needing `MaybeUninit`, even if it's a little less efficient.

The code is trivial, IMO the hard part is coming up with a clear name for this method in comparison to `recv_from`. Any better options?

fixes #223

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -493,6 +493,15 @@ impl Socket {
         self.recv_from_with_flags(buf, 0)
     }
 
+    /// Receives data from the socket. On success, returns the number of bytes
+    /// read and the address from whence the data came.
+    ///
+    /// Similar to [`recv_from`] except using a `&mut [u8]` initialised buffer.
+    pub fn recv_from_into_buf(&self, buf: &mut [u8]) -> io::Result<(usize, SockAddr)> {
+        let buf = unsafe { &mut *(buf as *mut [u8] as *mut [MaybeUninit<u8>]) };
+        self.recv_from(buf)
+    }
+
     /// Identical to [`recv_from`] but allows for specification of arbitrary
     /// flags to the underlying `recvfrom` call.
     ///


Thomasdezeeuw:
> It would be nice to have a way to use `recv_from` without needing `MaybeUninit`, even if it's a little less efficient.

Yeah it's a problem I struggled with as well and basically postponed until someone complained, and here we are... `TcpStream` has it easy as it just implemented the `io::Read` trait. I don't really want to add `&[u8]` variants of all functions, as there are many of them for `UdpSocket`, see `recv`, `recv_out_of_band`, etc, having a duplicate of all them using `&[u8]` is a bit much... 

> The code is trivial, IMO the hard part is coming up with a clear name for this method in comparison to `recv_from`. Any better options?

In my higher-level crate I've created a [Bytes](https://docs.rs/heph/0.3.0/heph/net/trait.Bytes.html) trait that helps with this, but it's not possible to add a sound implementation that does `&[u8]` -> `&[MaybeUninit<u8>]` as that would allow safe code to write uninit bytes to `&[u8]`.

I'm afraid that I don't have a good solution... Could use some more designs/opinions here.

keepsimple1:
Regarding the name, how about `recv_buf_from` ?  (I admit it is no longer following `recv_from_` prefix, but it also has its advantages, I think)

Thomasdezeeuw:
@fujiapple852 I think the `ReadBuf` API (std::io::ReadBuf) would be the way forward, but we'll have to wait until it stabilises: https://github.com/rust-lang/rust/issues/78485.

Thomasdezeeuw:
Closing in favour of #366.

PR #249, opened 2021-07-22
 Windows: Add support for Unix sockets

jstarks:
Newer versions of Windows support AF_UNIX stream sockets. This change
adds Windows support for the `SockAddr::unix` function and the
`Domain::UNIX` constant.

Since Unix sockets are now available on all tier1 platforms, this also
removes `all` feature requirement from the `SockAddr::unix` function.

### DIFF
--- a/src/sockaddr.rs
+++ b/src/sockaddr.rs
@@ -1,5 +1,6 @@
 use std::mem::{self, size_of, MaybeUninit};
 use std::net::{SocketAddr, SocketAddrV4, SocketAddrV6};
+use std::path::Path;
 use std::{fmt, io};
 
 #[cfg(windows)]
@@ -209,6 +210,16 @@ impl SockAddr {
             Some(SocketAddr::V6(addr)) => Some(addr),
             _ => None,
         }
+    }
+
+    /// Constructs a `SockAddr` with the family `AF_UNIX` and the provided path.
+    ///
+    /// Returns an error if the path is longer than `SUN_LEN`.
+    pub fn unix<P>(path: P) -> io::Result<SockAddr>
+    where
+        P: AsRef<Path>,
+    {
+        crate::sys::unix_sockaddr(path.as_ref())
     }
 }
 
--- a/src/sys/windows.rs
+++ b/src/sys/windows.rs
@@ -14,6 +14,7 @@ use std::net::{self, Ipv4Addr, Ipv6Addr, Shutdown};
 use std::os::windows::io::{
     AsRawSocket, AsSocket, BorrowedSocket, FromRawSocket, IntoRawSocket, OwnedSocket, RawSocket,
 };
+use std::path::Path;
 use std::sync::Once;
 use std::time::{Duration, Instant};
 use std::{process, ptr, slice};
@@ -41,8 +42,8 @@ pub(crate) const MSG_TRUNC: c_int = 0x01;
 // Used in `Domain`.
 pub(crate) const AF_INET: c_int = windows_sys::Win32::Networking::WinSock::AF_INET as c_int;
 pub(crate) const AF_INET6: c_int = windows_sys::Win32::Networking::WinSock::AF_INET6 as c_int;
-const AF_UNIX: c_int = windows_sys::Win32::Networking::WinSock::AF_UNIX as c_int;
-const AF_UNSPEC: c_int = windows_sys::Win32::Networking::WinSock::AF_UNSPEC as c_int;
+pub(crate) const AF_UNIX: c_int = windows_sys::Win32::Networking::WinSock::AF_UNIX as c_int;
+pub(crate) const AF_UNSPEC: c_int = windows_sys::Win32::Networking::WinSock::AF_UNSPEC as c_int;
 // Used in `Type`.
 pub(crate) const SOCK_STREAM: c_int = windows_sys::Win32::Networking::WinSock::SOCK_STREAM as c_int;
 pub(crate) const SOCK_DGRAM: c_int = windows_sys::Win32::Networking::WinSock::SOCK_DGRAM as c_int;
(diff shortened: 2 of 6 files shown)


src/sys/windows.rs
```diff
@@ -735,6 +736,52 @@ pub(crate) fn from_in6_addr(addr: in6_addr) -> Ipv6Addr {
     Ipv6Addr::from(*unsafe { addr.u.Byte() })
 }
 
+/// This type is not yet in winapi.
+#[repr(C)]
+#[allow(non_camel_case_types)]
+struct sockaddr_un {
```
Thomasdezeeuw, line 742: Note to self: pr to add this in winapi: https://github.com/retep998/winapi-rs/pull/968.

src/sys/windows.rs
```diff
@@ -735,6 +736,52 @@ pub(crate) fn from_in6_addr(addr: in6_addr) -> Ipv6Addr {
...
+pub(crate) fn unix_sockaddr(path: &Path) -> io::Result<SockAddr> {
+    unsafe {
+        SockAddr::init(|storage, len| {
+            // Safety: `SockAddr::init` zeros the address, which is a valid
+            // representation.
+            let storage: &mut crate::sys::sockaddr_un = unsafe { &mut *storage.cast() };
+            let len: &mut socklen_t = unsafe { &mut *len };
+
+            let bytes = path
+                .to_str()
```
Thomasdezeeuw, line 757: What we really need here is a `Path`/`OsStr` -> `&[u8]` method don't we? For Unix this already exists. For Windows all the way down this is a `Wtf8` which contains one field: `bytes: [u8]`, so it shouldn't be too hard to add.

src/sys/windows.rs
```diff
@@ -735,6 +736,52 @@ pub(crate) fn from_in6_addr(addr: in6_addr) -> Ipv6Addr {
...
+
+            if bytes.len() > storage.sun_path.len() {
+                return Err(io::Error::new(
+                    io::ErrorKind::InvalidInput,
+                    "path must be shorter than SUN_LEN",
+                ));
+            }
+
+            storage.sun_family = crate::sys::AF_UNIX as sa_family_t;
+            storage.sun_path[..bytes.len()].copy_from_slice(bytes);
```
Thomasdezeeuw, line 771: Can you mention `storage.sun_path` is zero filled and thus already null terminated, same as https://github.com/rust-lang/socket2/blob/1c67209c2eda5708b3093900a59befbc4a981b13/src/sys/unix.rs#L477-L478

src/sys/windows.rs
```diff
@@ -735,6 +736,52 @@ pub(crate) fn from_in6_addr(addr: in6_addr) -> Ipv6Addr {
...
+            let len: &mut socklen_t = unsafe { &mut *len };
+
+            let bytes = path
+                .to_str()
+                .ok_or_else(|| {
+                    io::Error::new(io::ErrorKind::InvalidInput, "path must be valid UTF-8")
+                })?
+                .as_bytes();
+
+            if bytes.len() > storage.sun_path.len() {
```
Thomasdezeeuw, line 763: This needs the same check as in https://github.com/rust-lang/socket2/blob/1c67209c2eda5708b3093900a59befbc4a981b13/src/sys/unix.rs#L460-L472

If it doesn't support abstract namespaces you can remove that bit. The correct code won't work if `bytes.len() == storage.sun_path.len()` as the string isn't null terminated then.

src/sys/windows.rs
```diff
@@ -735,6 +736,52 @@ pub(crate) fn from_in6_addr(addr: in6_addr) -> Ipv6Addr {
...
+                ));
+            }
+
+            storage.sun_family = crate::sys::AF_UNIX as sa_family_t;
+            storage.sun_path[..bytes.len()].copy_from_slice(bytes);
+
+            let base = storage as *const _ as usize;
+            let path = &storage.sun_path as *const _ as usize;
+            let sun_path_offset = path - base;
+            let length = sun_path_offset + bytes.len();
```
Thomasdezeeuw, line 776: I think this needs +1 for the null terminator, see https://github.com/rust-lang/socket2/blob/1c67209c2eda5708b3093900a59befbc4a981b13/src/sys/unix.rs#L490-L495

Thomasdezeeuw (changes requested):

src/sys/windows.rs
jstarks: The Windows implementation doesn't require null termination as long as the length is set appropriately. I can include this for consistency if you'd like, but I don't think it's strictly necessary.

src/sys/windows.rs
jstarks: Windows should allow the full sun_path to be filled out without a null terminator.

src/sys/windows.rs
jstarks: Internally Windows uses a 16-bit representation and converts sun_path to it. Unfortunately, this conversion will reject strings with unmatched Unicode surrogates, meaning that it will only accept UTF-8 strings, not the WTF-8 superset, even though the 16-bit representation for Windows paths _does_ allow unmatched surrogates.

So even if this method existed and we used it here, it would just defer the failure to the point of the bind/connect call.

Thomasdezeeuw: > So even if this method existed and we used it here, it would just defer the failure to the point of the bind/connect call.

That's fine we don't check if the path is valid for Unix either (except for correct usage, i.e. null terminated), we just pass it along to the OS.

src/sys/windows.rs
Thomasdezeeuw: In that case can you add a comment stating that and referring to the Windows docs? I the only docs I read about it said it was Linux compatible (which it still is, but the not null-terminated usage isn't as that is invalid on Unix).

src/sys/windows.rs
Thomasdezeeuw: See comment above, not necessary.

src/sys/windows.rs
jstarks: windows-sys has this definition so I've removed the private one.

src/sys/windows.rs
jstarks: I used to manage the team that implemented this in Windows, but it appears that we didn't document this behavior. I'll keep the null terminator requirement for now and we can relax it once I get the documentation fixed.

src/sys/windows.rs
jstarks: I've clarified this in a comment.

src/sys/windows.rs
jstarks: Done.

jstarks:
I've finally addressed your comments and have rebased for 0.5.

notgull:
Just to clarify: if this is called on a Windows platform that doesn't support UDS, it just returns an error, right?

jstarks:
Right, it will fail with `WSAEAFNOSUPPORT`.

src/sockaddr.rs
```diff
@@ -210,6 +211,18 @@ impl SockAddr {
             _ => None,
         }
     }
+
+    /// Constructs a `SockAddr` with the family `AF_UNIX` and the provided path.
+    ///
+    /// # Failure
+    ///
```
Thomasdezeeuw, line 218: Nit: we stopped using this.

```suggestion
```

src/sys/windows.rs
```diff
@@ -774,6 +775,53 @@ pub(crate) fn to_mreqn(
     }
 }
 
+#[allow(unused_unsafe)] // TODO: replace with `unsafe_op_in_unsafe_fn` once stable.
+pub(crate) fn unix_sockaddr(path: &Path) -> io::Result<SockAddr> {
+    unsafe {
+        SockAddr::try_init(|storage, len| {
```
Thomasdezeeuw, line 781: We've change our `SockAddr` constructors usage a bit since this was written, can you change it to use `SockAddr::new`, e.g: https://github.com/rust-lang/socket2/blob/eaa6300e13d7ed421a3df2fe0149b7a54956b84c/src/sys/unix.rs#L540-L552

src/sys/windows.rs
```diff
@@ -774,6 +775,53 @@ pub(crate) fn to_mreqn(
...
+            // buffer, this could be used directly, relying on Windows to
+            // validate the path, but Rust hides this implementation detail.
+            let bytes = path
+                .to_str()
+                .ok_or_else(|| {
+                    io::Error::new(io::ErrorKind::InvalidInput, "path must be valid UTF-8")
+                })?
+                .as_bytes();
+
+            // Windows appears to allow non-null-terminated paths, but this is
```
Thomasdezeeuw, line 799: I don't remember where I read this, but the a person that was part of the team that wrote this code said it was an oversight in the docs and they would make an internal issue to update the docs to guarantee this. So, we should probably keep the code as is for the moment and open an issue to allow it in the future (in case some really needs that one additional byte).

src/sys/windows.rs
Thomasdezeeuw: We should also open an issue with the Rust repo to allow use access to the underlying bytes.

src/sys/windows.rs
jstarks: I was modeling this after the implementation in `unix.rs`. Would you like me to change both?

src/sys/windows.rs
jstarks: Right, that was my intention (I'm also the one who wrote the comment about it being an oversight in the documentation). I can open an issue and reference it here.

src/sys/windows.rs
jstarks: I found an existing draft PR that discusses this, I'll link it. https://github.com/rust-lang/rust/pull/95290

src/sys/windows.rs
jstarks: #331.

src/sys/windows.rs
Thomasdezeeuw: Yes please.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @jstarks.

Issue #85, opened 2020-06-17
Haiku test failure. keepalive / connect_timeout/unbound


kallisti5:
Just reporting a test failure on Haiku. Overall it's pretty close!

master as of 32969e53c75849e4047233d8b55f551a8645b21d

```
/Data/socket2-rs> cargo test
  Downloaded remove_dir_all v0.5.3
  Downloaded rand v0.4.6
  Downloaded tempdir v0.3.7
  Downloaded 3 crates (97.1 KB) in 0.93s
   Compiling remove_dir_all v0.5.3
   Compiling rand v0.4.6
   Compiling tempdir v0.3.7
   Compiling socket2 v0.3.12 (/Data/socket2-rs)
    Finished test [unoptimized + debuginfo] target(s) in 5.27s
     Running target/debug/deps/socket2-d90163a26b657973

running 13 tests
test socket::test::connect_timeout_unbound ... FAILED
test socket::test::connect_timeout_unrouteable ... ok
test socket::test::connect_timeout_valid ... ok
test socket::test::keepalive ... FAILED
test socket::test::nodelay ... ok
test socket::test::tcp ... ok
test sys::test_ip ... ok
test tests::domain_fmt_debug ... ok
test tests::domain_for_address ... ok
test tests::protocol_fmt_debug ... ok
test tests::socket_address_ipv4 ... ok
test tests::socket_address_ipv6 ... ok
test tests::type_fmt_debug ... ok

failures:

---- socket::test::connect_timeout_unbound stdout ----
thread 'main' panicked at 'unexpected success', src/socket.rs:925:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- socket::test::keepalive stdout ----
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: -2147483643, kind: InvalidInput, message: "Invalid Argument" }', src/socket.rs:985:9


failures:
    socket::test::connect_timeout_unbound
    socket::test::keepalive

test result: FAILED. 11 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '--lib'
```

Thomasdezeeuw:
@alexcrichton is there anyone we can ping for Haiku support?

alexcrichton:
I am unaware of someone myself.

Thomasdezeeuw:
[...]

PR #251, opened 2021-07-27
Fix TCP keepalive handling on Haiku and OpenBSD

link2xt:
On Haiku and OpenBSD don't call `setsockopt(fd, IPPROTO_TCP,
SO_KEEPALIVE, secs)`, it is incorrect because `SO_KEEPALIVE` belongs
to `SOL_SOCKET` layer, because it accepts a boolean instead of the
number of seconds, and because there is no way to set keepalive idle
time per socket on these operating systems.

On Haiku and OpenBSD it is only possible to enable and disable
keepalives, and it is already done with `setsockopt(fd, SOL_SOCKET,
SO_KEEPALIVE, 1)` in a portable way in `Socket.set_keepalive()`
defined in `src/socket.rs`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -340,10 +340,11 @@ impl TcpKeepalive {
     /// Set the amount of time after which TCP keepalive probes will be sent on
     /// idle connections.
     ///
-    /// This will set the value of `SO_KEEPALIVE` on OpenBSD and Haiku,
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
Thanks for looking into this. Does this fix #85?

Thomasdezeeuw (approved):

link2xt:
> Does this fix #85?

No, there are still two other test failures.

PR #256, opened 2021-07-28
Fix compilation on Haiku and OpenBSD with --features all

### LINKED ISSUE #255: socket2 0.4.1 does not compile on Haiku with feature `all`
The problem is that since merging of #251 on this line `KEEPALIVE_TIME` is not defined on Haiku and OpenBSD:
https://github.com/rust-lang/socket2/blob/877a789982a87bb487381ada36f15c0786fa221b/src/sys/unix.rs#L871

link2xt:
`keepalive_time()` is not available on these platforms

Fixes #255

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -1404,11 +1404,22 @@ impl Socket {
 impl Socket {
     /// Get the value of the `TCP_KEEPIDLE` option on this socket.
     ///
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @link2xt

PR #258, opened 2021-08-14
Use correct socket option level and name to get/set freebind flag

### LINKED ISSUE #257: freebind option getter/setter actually get/set the reuseport flag
We are passing an incorrect socket option level (`SOL_SOCKET` instead of `SOL_IP`/`SOL_IPV6`) and `SO_REUSEPORT` happens to have the same value as `IP_FREEBIND` on Linux:

```
$ strace -fe getsockopt,setsockopt cargo test -q --all-features freebind
strace: Process 496571 attached
[pid 496571] --- SIGUSR1 {si_signo=SIGUSR1, si_code=SI_TKILL, si_pid=496570, si_uid=1000} ---
[pid 496571] +++ exited with 0 +++
strace: Process 496572 attached

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 4 filtered out; finished in 0.00s

[pid 496572] +++ exited with 0 +++
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=496572, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---
strace: Process 496573 attached

running 1 test
strace: Process 496574 attached
[pid 496574] getsockopt(3, SOL_SOCKET, SO_REUSEPORT, [0], [4]) = 0
[pid 496574] setsockopt(3, SOL_SOCKET, SO_REUSEPORT, [1], 4) = 0
[pid 496574] getsockopt(3, SOL_SOCKET, SO_REUSEPORT, [1], [4]) = 0
[pid 496574] getsockopt(4, SOL_SOCKET, SO_REUSEPORT, [0], [4]) = 0
[pid 496574] setsockopt(4, SOL_SOCKET, SO_REUSEPORT, [1], 4) = 0
[pid 496574] getsockopt(4, SOL_SOCKET, SO_REUSEPORT, [1], [4]) = 0
[pid 496574] +++ exited with 0 +++
.
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 52 filtered out; finished in 0.01s

[pid 496573] +++ exited with 0 +++
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=496573, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---
+++ exited with 0 +++
```

jsitnicki:
`freebind()`/`set_freebind()` are currently getting/setting the reuseport flag because we are passing an incorrect socket level - `SOL_SOCKET` instead of `SOL_IP`, and `SO_REUSEPORT` happens to have the same value as `IP_FREEBIND` on Linux:
```
  $ git -C ~/src/linux grep -h 'define \(SO_REUSEPORT\|IP_FREEBIND\)' -- include/uapi
  #define SO_REUSEPORT    15
  #define IP_FREEBIND     15
```
Fix it by using `(SOL_IP, IP_FREEBIND)` and `(SOL_IPV6, IPV6_FREEBIND)` socket option (level, name) to get/set the freebind flag, for INET and INET6 sockets, respectively.

Alternatively, `(SOL_IP, IP_FREEBIND)` could be used for both IPv4 and IPv6 because Linux kernel underneath toggles a flag that is shared by v4 and v6 network stack. However, this is an implementation detail, and arguably not what would be expected by the user when looking at strace output.

Fixes #257.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -1478,8 +1478,8 @@ impl crate::Socket {
     )]
     pub fn freebind(&self) -> io::Result<bool> {
         unsafe {
-            getsockopt::<c_int>(self.as_raw(), libc::SOL_SOCKET, libc::IP_FREEBIND)
-                .map(|reuse| reuse != 0)
+            getsockopt::<c_int>(self.as_raw(), libc::SOL_IP, libc::IP_FREEBIND)
+                .map(|freebind| freebind != 0)
         }
     }
 
@@ -1501,13 +1501,13 @@ impl crate::Socket {
             any(target_os = "android", target_os = "fuchsia", target_os = "linux")
         )))
     )]
-    pub fn set_freebind(&self, reuse: bool) -> io::Result<()> {
+    pub fn set_freebind(&self, freebind: bool) -> io::Result<()> {
         unsafe {
             setsockopt(
                 self.as_raw(),
-                libc::SOL_SOCKET,
+                libc::SOL_IP,
                 libc::IP_FREEBIND,
-                reuse as c_int,
+                freebind as c_int,
             )
         }
     }
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -1512,6 +1512,71 @@ impl crate::Socket {
         }
     }
 
+    /// Get the value of the `IPV6_FREEBIND` option on this socket.
+    ///
+    /// This is an IPv6 counterpart of `IP_FREEBIND` socket option on
+    /// Android/Linux. For more information about this option, see
+    /// [`set_freebind`].
+    ///
+    /// [`set_freebind`]: crate::Socket::set_freebind
+    #[cfg(all(feature = "all", any(target_os = "android", target_os = "linux")))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(feature = "all", any(target_os = "android", target_os = "linux"))))
+    )]
+    pub fn freebind_ipv6(&self) -> io::Result<bool> {
+        unsafe {
+            getsockopt::<c_int>(self.as_raw(), libc::SOL_IPV6, libc::IPV6_FREEBIND)
+                .map(|freebind| freebind != 0)
+        }
+    }
+
+    /// Set value for the `IPV6_FREEBIND` option on this socket.
+    ///
+    /// This is an IPv6 counterpart of `IP_FREEBIND` socket option on
+    /// Android/Linux. For more information about this option, see
+    /// [`set_freebind`].
+    ///
+    /// [`set_freebind`]: crate::Socket::set_freebind
+    ///
+    /// # Examples
+    ///
+    /// On Linux:
+    ///
+    /// ```
+    /// use socket2::{Domain, Socket, Type};
+    /// use std::io::{self, Error, ErrorKind};
+    ///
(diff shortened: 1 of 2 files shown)


jsitnicki:
Interesting, CI failures show that `libc::IPV6_FREEBIND` doesn't exist on x86_64-fuchsia.

Let me dig around. Maybe implementation needs to be split up for Linux and Fuchsia.

Thomasdezeeuw (commented):
@jsitnicki I think you're right, but man these are poorly documented... Neither https://man7.org/linux/man-pages/man7/ip.7.html or https://man7.org/linux/man-pages/man7/ipv6.7.html mentions `IPV6_FREEBIND`. I can only conclude `SOL_IPV6` is needed and that `IPV6_FREEBIND` even exists is by looking a the Linux source, at https://github.com/torvalds/linux/blob/5bfc75d92efd494db37f5c4c173d3639d4772966/net/ipv6/ipv6_sockglue.c#L638 and 
https://github.com/torvalds/linux/blob/8096acd7442e613fad0354fc8dfdb2003cceea0b/net/mptcp/sockopt.c#L456-L480.

Nice find!

But now the solution. I don't think I want the hidden call to `domain`, socket2 tries to map methods directly to system calls. Can you instead fix the current functions by changing `SOL_SOCKET` to `SOL_IP`. And add a new variant `(set)_freebind_ipv6`? It would be nice if you could add the code that checks `domain` as an example.

Thomasdezeeuw:
Oh, as for the missing constant on Fuschia I'll add it to libc.

Thomasdezeeuw:
CI blocked on https://github.com/rust-lang/libc/pull/2341.

Thomasdezeeuw:
@jsitnicki Fuchsia doesn't seem to support `IPV6_FREEBIND`, so you can remove support for the v6 version from it.

jsitnicki:
Thanks for quick feedback @Thomasdezeeuw !

> Neither https://man7.org/linux/man-pages/man7/ip.7.html or https://man7.org/linux/man-pages/man7/ipv6.7.html mentions IPV6_FREEBIND.

Right, unfortunately there are quite a few IPv6 socket options present on Linux that are not documented by the ipv6(7) man page (yet):

https://elixir.bootlin.com/linux/v5.13.10/source/include/uapi/linux/in6.h#L152

> But now the solution. I don't think I want the hidden call to domain, socket2 tries to map methods directly to system calls. 

That is fair. I was wondering how "smart" the `freebind`/`set_freebind` wrappers should be. I will split it up as suggested.

jsitnicki:
v2:
* fixed `freebind`/`set_freebind` to use `SOL_IP` socket level,
* added `freebind_ipv6`/`set_freebind_ipv6` to access (`SOL_IPV6`, `IPV6_FREEBIND`) socket option on Linux platforms

jsitnicki:
> It would be nice if you could add the code that checks domain as an example.

One moment. Forgot to add an example.

Thomasdezeeuw (approved):

jsitnicki:
v3:
* added an example for `set_freebind_ipv6`

jsitnicki:
v4:
* fix build failure against 1.46 due to missing `std::io::ErrorKind::Unsupported` variant.

Thomasdezeeuw:
Thanks @jsitnicki.

Issue #259, opened 2021-08-16
Use of set_only_v6() in example code


cmcqueen:
In the [example](https://docs.rs/socket2/0.4.1/socket2/#examples) code in `lib.rs`, it has these lines:

    socket.bind(&address.into())?;
    socket.set_only_v6(false)?;

Isn't it necessary for the call to `set_only_v6()` to be *before* the `bind()`?

Thomasdezeeuw:
Yes. Want to send a pr?

PR #260, opened 2021-08-17
Example doc comment: fix usage of socket.set_only_v6()

cmcqueen:
The call to `socket.set_only_v6()` needs to be *before* the call to
`socket.bind()`.

This is for issue #259.


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @cmcqueen.

PR #263, opened 2021-09-03
Add KEEPALIVE_TIME for haiku and openbsd (compilation was broken)

epilys:
Compilation on openbsd was broken since KEEPALIVE_TIME was not defined anywhere.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -135,6 +135,8 @@ pub(crate) use libc::{TCP_KEEPCNT, TCP_KEEPINTVL};
 // See this type in the Windows file.
 pub(crate) type Bool = c_int;
 
+#[cfg(any(target_os = "haiku", target_os = "openbsd"))]
+use libc::SO_KEEPALIVE as KEEPALIVE_TIME;
 #[cfg(target_vendor = "apple")]
 use libc::TCP_KEEPALIVE as KEEPALIVE_TIME;
 #[cfg(not(any(target_vendor = "apple", target_os = "haiku", target_os = "openbsd")))]


Thomasdezeeuw:
`SO_KEEPALIVE` is not the same as `KEEPALIVE_TIME`, also see
https://github.com/rust-lang/socket2/pull/251.

epilys:
Oh. I read the [manpage](https://man.openbsd.org/getsockopt.2) and it doesn't mention something like that. Would returning an error be acceptable? Otherwise it won't build at all.

epilys:
And [on sysctl(3)](https://man.openbsd.org/OpenBSD-5.9/man3/sysctl.3#tcp.always_keepalive) it mentions KEEPALIVE set on TCP sockets.

Thomasdezeeuw:
> Oh. I read the [manpage](https://man.openbsd.org/getsockopt.2) and it doesn't mention something like that.

`KEEPALIVE_TIME` is `TCP_KEEPALIVE`, thus on the TCP socket level, while `SO_KEEPALIVE` is on the more general socket level (`SO`).

>  Would returning an error be acceptable? Otherwise it won't build at all.

I'm afraid not, but we can remove the function from OpenBSD by using something like:

```rust
#[cfg(not(target_os = "openbsd"))]
```

on `keepalive_time` and `set_keepalive_time`.

> And [on sysctl(3)](https://man.openbsd.org/OpenBSD-5.9/man3/sysctl.3#tcp.always_keepalive) it mentions KEEPALIVE set on TCP sockets.

That is a boolean to enable keepalive, while `KEEPALIVE_TIME` sets a duration, see [`TCP_KEEPIDLE`](https://man7.org/linux/man-pages/man7/tcp.7.html) for Linux or `TCP_KEEPALIVE` for macOS.

PR #267, opened 2021-09-17
Add compile_error! for unsupported targets

Thomasdezeeuw:
Although it still spits out a bunch of errors at least the first one
will be informative about the root cause.

https://github.com/rust-lang/rust/issues/68838 tracks a version that
suppresses subsequent errors.

/cc @twitchax not a huge improvement, but still an improvement.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,6 +126,9 @@ mod sys;
 #[path = "sys/windows.rs"]
 mod sys;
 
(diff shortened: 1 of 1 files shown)


twitchax:
Ahhh, very nice!

Honestly, I love the fact that the Rust community puts such a huge emphasis on informative errors.  I think this is going to save multiple people multiple minutes over the next forever. 😄

PR #273, opened 2021-11-01
Add support to attach filter onto socket on Linux

hhyhhy:
Hi

This patch add support to attach filter onto socket on Linux.
See also: [Linux Socket Filtering aka Berkeley Packet Filter (BPF)](https://www.kernel.org/doc/html/v5.12/networking/filter.html)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -131,9 +131,6 @@ compile_error!("Socket2 doesn't support the compile target");
 
 use sys::c_int;
 
-#[cfg(all(feature = "all", target_os = "linux"))]
-pub use sys::SockFilter;
-
 pub use sockaddr::SockAddr;
 pub use socket::Socket;
 pub use sockref::SockRef;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -131,6 +131,9 @@ compile_error!("Socket2 doesn't support the compile target");
 
 use sys::c_int;
 
+#[cfg(all(feature = "all", target_os = "linux"))]
+pub use sys::SockFilter;
+
 pub use sockaddr::SockAddr;
 pub use socket::Socket;
 pub use sockref::SockRef;
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -1781,10 +1781,10 @@ impl crate::Socket {
     ///
     /// For more information about this option, see [filter](https://www.kernel.org/doc/html/v5.12/networking/filter.html)
     #[cfg(all(feature = "all", any(target_os = "linux", target_os = "android")))]
-    pub fn attach_filter(&self, filters: &mut [libc::sock_filter]) -> io::Result<()> {
+    pub fn attach_filter(&self, filters: &[libc::sock_filter]) -> io::Result<()> {
         let prog = libc::sock_fprog {
             len: filters.len() as u16,
-            filter: filters.as_mut_ptr(),
+            filter: filters.as_ptr() as *mut _,
         };
 
         unsafe {
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -10,8 +10,6 @@ use std::cmp::min;
 #[cfg(not(target_os = "redox"))]
 use std::io::IoSlice;
 use std::marker::PhantomData;
-#[cfg(all(feature = "all", target_os = "linux"))]
-use std::mem::forget;
 use std::mem::{self, size_of, MaybeUninit};
 use std::net::Shutdown;
 use std::net::{Ipv4Addr, Ipv6Addr};
@@ -56,9 +54,6 @@ use libc::{c_void, in6_addr, in_addr};
 #[cfg(not(target_os = "redox"))]
 use crate::RecvFlags;
 use crate::{Domain, Protocol, SockAddr, TcpKeepalive, Type};
-
-#[cfg(all(feature = "all", target_os = "linux"))]
-pub(crate) use libc::{c_uchar, c_uint, c_ushort};
 
 pub(crate) use libc::c_int;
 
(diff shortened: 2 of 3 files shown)


src/sys/unix.rs
```diff
@@ -200,6 +205,31 @@ type IovLen = usize;
 ))]
 type IovLen = c_int;
 
+/// Required for SO_ATTACH_FILTER.
+#[cfg(all(feature = "all", target_os = "linux"))]
+#[derive(Debug)]
+#[repr(C)]
+struct SockFprog {
```
Thomasdezeeuw, line 212: Please us `libc` https://docs.rs/libc/0.2.106/libc/struct.sock_fprog.html.

src/sys/unix.rs
```diff
@@ -200,6 +205,31 @@ type IovLen = usize;
...
+    ///  Number of filter blocks
+    pub len: c_ushort,
+    pub filter: *mut SockFilter,
+}
+
+/// Linux Filter block
+#[cfg(all(feature = "all", target_os = "linux"))]
+#[derive(Debug)]
+#[repr(C)]
+pub struct SockFilter {
```
Thomasdezeeuw, line 222: Please us `libc` https://docs.rs/libc/0.2.106/libc/struct.sock_filter.html.

src/lib.rs
```diff
@@ -131,6 +131,9 @@ compile_error!("Socket2 doesn't support the compile target");
 
 use sys::c_int;
 
+#[cfg(all(feature = "all", target_os = "linux"))]
+pub use sys::SockFilter;
```
Thomasdezeeuw, line 135: Don't expose this.

src/sys/unix.rs
```diff
@@ -1773,6 +1803,41 @@ impl crate::Socket {
...
         }
     }
+
+    /// Attach Berkeley Packet Filter(BPF) on this socket.
+    ///
+    /// BPF allows a user-space program to attach a filter onto any socket
+    /// and allow or disallow certain types of data to come through the socket.
+    ///
+    /// For more information about this option, see [filter](https://www.kernel.org/doc/html/v5.12/networking/filter.html)
+    #[cfg(all(feature = "all", target_os = "linux"))]
```
Thomasdezeeuw, line 1813: Does Android support this?

src/sys/unix.rs
```diff
@@ -1773,6 +1803,41 @@ impl crate::Socket {
...
+    pub fn attach_filter(&self, mut filters: Vec<SockFilter>) -> io::Result<()> {
```
Thomasdezeeuw, line 1814: We don't need ownership of the `filter`s, use `&[libc::sock_filter]` instead and we can create the `libc::sock_fprog` structure based on that.

src/sys/unix.rs
```diff
@@ -1773,6 +1803,41 @@ impl crate::Socket {
...
+            filter: filters.as_mut_ptr(),
+        };
+        forget(filters);
+
+        unsafe {
+            setsockopt(
+                self.as_raw(),
+                libc::SOL_SOCKET,
+                libc::SO_ATTACH_FILTER,
+                prog,
```
Thomasdezeeuw, line 1829: Pretty sure this needs to be `&prog`.

Thomasdezeeuw (changes requested):

src/sys/unix.rs
hhyhhy: The payload has been converted to a pointer at: 

https://github.com/rust-lang/socket2/blob/5c7b03f61ef9ddb21233a18f643d6b9367429c74/src/sys/unix.rs#L963

I tried to change to `&prog` and it did not run successfully as before.

src/sys/unix.rs
hhyhhy: According to the filter [documentation](https://android.googlesource.com/kernel/msm/+/refs/tags/android-4.4w_r10/Documentation/networking/filter.txt) in android kernel, it should be supported.

I will fix it.

src/sys/unix.rs
```diff
@@ -1773,6 +1773,37 @@ impl crate::Socket {
...
     }
+
+    /// Attach Berkeley Packet Filter(BPF) on this socket.
+    ///
+    /// BPF allows a user-space program to attach a filter onto any socket
+    /// and allow or disallow certain types of data to come through the socket.
+    ///
+    /// For more information about this option, see [filter](https://www.kernel.org/doc/html/v5.12/networking/filter.html)
+    #[cfg(all(feature = "all", any(target_os = "linux", target_os = "android")))]
+    pub fn attach_filter(&self, filters: &mut [libc::sock_filter]) -> io::Result<()> {
```
Thomasdezeeuw, line 1784: We don't need a mutable reference I think? I assume the kernel is only reading the values.

You can use `filters.as_ptr() as *mut _`.

src/sys/unix.rs
Thomasdezeeuw: You're right, I should really try to remember the code I wrote :)

Thomasdezeeuw (commented):
Final small change.

Thomasdezeeuw:
It seems aarch64-linux-android doesn't have `libc::SO_ATTACH_FILTER`. We should add that to the libc crate.

hhyhhy:
I have created a [pr](https://github.com/rust-lang/libc/pull/2506) for libc.

Thomasdezeeuw:
I think this now depends on https://github.com/rust-lang/libc/pull/2506.

Thomasdezeeuw:
@hhyhhy if you can update to  libc v0.2.107 this can be merged 👍

hhyhhy:
It seems that the problem on the Android system has been solved.

Thomasdezeeuw:
Thanks @hhyhhy

PR #286, opened 2022-01-17
Add support for the socket options TCP_CORK, TCP_QUICKACK and TCP_THIN_LINEAR_TIMEOUTS

### LINKED ISSUE #285: Support TCP_CORK, TCP_THIN_LINEAR_TIMEOUTS and TCP_QUICKACK
Hallo,

for my project I need to use the socket options TCP_CORK, TCP_THIN_LINEAR_TIMEOUTS and TCP_QUICKACK and I'd like to implement them for socket2. As far as I know these options are Linux only.

Disclaimer: I am quite new to Rust and this is my first PR for a Rust project.

I am not sure where to implement this stuff in https://github.com/rust-lang/socket2/blob/master/src/sys/unix.rs and https://github.com/rust-lang/socket2/blob/master/src/socket.rs. I found a **unix only** section but I don't know if these options are available for other Unix-like systems as well. 

My first though is to put the option flags into https://github.com/rust-lang/socket2/blob/master/src/sys/unix.rs#L1004 and mark them with `#[cfg(target_os = "linux")]` and put the implementation into https://github.com/rust-lang/socket2/blob/master/src/socket.rs#L1445 and mark them also with the linux target.

The implementation itself would be more or less a copy/paste of nodelay/set_nodelay. 

What are your thoughts about it? 

Thanks

eddi0815:
Hi,

this is my first try and I ask for a review. I encounter one problem:

On my system TCP_QUICKACK is enabled by default. So the test fails. I could change the test, so that it deactivates TCP_QUICKACK, but on a different system it might be disabled by default. So that this test is flaky at the moment. 

The test could be changed that it checks if TCP_QUICKACK is enabled or not and then activate/deactivate the flag.

Closes #285

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -1385,6 +1385,60 @@ impl crate::Socket {
         }
     }
 
+    /// Get the value of the `TCP_THIN_LINEAR_TIMEOUTS` option on this socket.
+    ///
+    /// For more information about this option, see [`set_thin_linear_timeouts`].
+    ///
+    /// [`set_thin_linear_timeouts`]: Socket::set_thin_linear_timeouts
+    #[cfg(all(
+        feature = "all",
+        any(target_os = "android", target_os = "fuchsia", target_os = "linux")
+    ))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(
+            feature = "all",
+            any(target_os = "android", target_os = "fuchsia", target_os = "linux")
+        )))
+    )]
+    pub fn thin_linear_timeouts(&self) -> io::Result<bool> {
+        unsafe {
+            getsockopt::<Bool>(
+                self.as_raw(),
+                libc::IPPROTO_TCP,
+                libc::TCP_THIN_LINEAR_TIMEOUTS,
+            )
+            .map(|timeouts| timeouts != 0)
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


src/sys/unix.rs
```diff
@@ -1283,6 +1283,158 @@ impl crate::Socket {
...
+            getsockopt::<Bool>(self.as_raw(), libc::IPPROTO_TCP, libc::TCP_CORK)
+                .map(|cork| cork != 0)
+        }
+    }
+
+    /// Set the value of the `TCP_CORK` option on this socket.
+    ///
+    /// If set, don't send out partial frames. All queued partial frames are
+    /// sent when the option is cleared again. There is a 200 millisecond ceiling on
+    /// the time for which output is corked by TCP_CORK. If this ceiling is reached,
```
Thomasdezeeuw, line 1313: Nit: can you put `TCP_CORK` in backticks ("`")?

Thomasdezeeuw (commented):
Apart from one small comment the code looks good to me.

The CI failed however:
* You should run `cargo fmt` on the code.
* The other failures are due to the quickack being true by default, could you change that to false.

Finally could you split up the three new features into three commits? If not I can also do this.

src/sys/unix.rs
eddi0815: Done

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @eddi0815!

PR #288, opened 2022-01-22
Modernise license format in Cargo.toml

dimbleby:
https://doc.rust-lang.org/cargo/reference/manifest.html#slash

> Previously multiple licenses could be separated with a /, but that usage is deprecated.

removes ambiguity about whether AND or OR is intended, I guess.


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @dimbleby.

PR #291, opened 2022-01-28
Fix OpenBSD build

niklasha:
Rust libc does not provide ip_mreqn for openbsd, so add it to the cfg exclusion sets where necessary.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -1164,6 +1164,7 @@ impl Socket {
         target_os = "haiku",
         target_os = "illumos",
         target_os = "netbsd",
+        target_os = "openbsd",
         target_os = "redox",
         target_os = "solaris",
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw (approved):

Thomasdezeeuw:
According to https://man.openbsd.org/ip.4 `ip_mreqn` does exists on OpenBSD, so maybe the better route is to add it to libc?

niklasha:
> According to https://man.openbsd.org/ip.4 `ip_mreqn` does exists on OpenBSD, so maybe the better route is to add it to libc?

Yes I am aware of that.  However I was in a hurry, and had no time to check if the ip_mreqn use there, is compatible with libc's.  I might get back to it later.

PR #294, opened 2022-02-17
keepalive: allow retries & interval on tier 1s

fulara:
Probably there was a reason to it, but I am not sure why the tier ones were gated behind `all` feature here? 

I'd like to fully configure keepalive on linux, using nice api provided here, but currently I can't without enabling `all` feature.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -371,29 +371,35 @@ impl TcpKeepalive {
     ///
     /// Some platforms specify this value in seconds, so sub-second
     /// specifications may be omitted.
-    #[cfg(all(
-        feature = "all",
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
I don't remember exactly why we put these behind the `all` flag.

Maybe it's because the getters aren't available on pre Windows 10?
https://github.com/rust-lang/socket2/blob/c288e26816b06829e58a676ef4c5d2559be043b9/src/socket.rs#L1551-L1623

Thomasdezeeuw:
Closing in favour of https://github.com/rust-lang/socket2/pull/390.

PR #295, opened 2022-03-01
Changelog: UB fix in correct version

Kixunil:
Layout related UB was fixed in 0.3.16 but it was reported in 0.4.0 which would lead people to believe they have broken dependency even if they don't.
Moving it to appropriate version avoids confusion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -169,7 +169,6 @@
 * Use correct inmutable references in `Socket::send_with_flags` and
   `Socket::send_out_of_band`.
 * Use `IPPROTO_IPV6` in `Socket::join_multicast_v6` on Windows.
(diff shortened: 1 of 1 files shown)


Kixunil:
CI failure is unrelated.

Thomasdezeeuw:
Socket didn't keep a changelog before v0.4, I started doing it when I took over maintenance. I'm not sure I want to start now for v0.3. I know you're coming from https://github.com/rust-lang/rust/pull/78802, all broken version have been long yanked.

Kixunil:
Surely, I don't expect you to add more entries. This one seems pretty important to me. Do you see any harm adding just this one?

Thomasdezeeuw:
Fair enough. Thanks @Kixunil.

PR #299, opened 2022-04-22
Added the recv_tos socket option

JoNil:


src/sys/unix.rs
```diff
@@ -83,6 +83,13 @@ pub(crate) use libc::IP_HDRINCL;
     target_os = "solaris",
     target_os = "illumos",
 )))]
+pub(crate) use libc::IP_RECVTOS;
```
Thomasdezeeuw, line 86: I think this `cfg` is the same as the one for `IP_TOS`, can you merge the two.

Thomasdezeeuw (changes requested):
The addition sounds good, but this needs tests and Windows support (see https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options).

JoNil:
I didn't find the IP_RECVTOS symbol in winapi-rs. Does anyone know if it is supported there and is just not exposed?

Thomasdezeeuw:
> I didn't find the IP_RECVTOS symbol in winapi-rs. Does anyone know if it is supported there and is just not exposed?

I can't find it either.  Perhaps we should move to `windows_sys` (it has it: https://docs.rs/windows-sys/latest/windows_sys/Win32/Networking/WinSock/constant.IP_RECVTOS.html), but that's a larger topic. For now you can remove Window support and we'll open issue to add it later.

JoNil:
In the last commit i added windows the the cfg so that it is not included :)

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @JoNil.

PR #305, opened 2022-05-08
Relax windows-sys dependency version

atouchet:
Is setting equals 0.36 exactly necessary here?


Thomasdezeeuw:
I think I copied it over from Mio, oops. Thanks @atouchet.

Thomasdezeeuw (approved):

Issue #310, opened 2022-05-13
API to clone/copy a SockAddr?


nathaniel-bennett:
Is there any way to create a new copy of a `SockAddr` from an existing one? I've searched through documentation pretty extensively, but there doesn't appear to be any API to do it as it stands.

It looks like `sockaddr_storage` and `socklen_t` have both `Copy` and `Clone` traits implemented for them, and in its current form `SockAddr` is composed of just those two values, so unless there's some compatibility reason for not adding them it seems like it would be pretty straightforward to add `Copy`/`Clone` traits to `SockAddr`. I would be happy to help out with implementing something like this if wanted :).

Thomasdezeeuw:
A pr to add `Clone` would be accept. I'm sure about `Copy`, it's a little big I think.

Thomasdezeeuw:
Fixed in #311.

PR #311, opened 2022-05-13
Clone trait for SockAddr

nathaniel-bennett:
Fix for #310 - adds the `Clone` trait to `SockAddr` to address certain use cases.

### DIFF
--- a/src/sockaddr.rs
+++ b/src/sockaddr.rs
@@ -1,6 +1,6 @@
 use std::mem::{self, size_of, MaybeUninit};
 use std::net::{SocketAddr, SocketAddrV4, SocketAddrV6};
-use std::{clone, fmt, io};
+use std::{fmt, io};
 
 #[cfg(windows)]
 use windows_sys::Win32::Networking::WinSock::SOCKADDR_IN6_0;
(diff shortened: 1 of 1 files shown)


src/sockaddr.rs
```diff
@@ -305,6 +305,15 @@ impl fmt::Debug for SockAddr {
     }
 }
 
+impl clone::Clone for SockAddr {
```
Thomasdezeeuw, line 308: I think we can just derive using `#[derive(Clone)]`.

nathaniel-bennett: Right--replaced that :+1:

src/sockaddr.rs
```diff
@@ -14,6 +14,7 @@ use crate::sys::{
 ///
 /// `SockAddr`s may be constructed directly to and from the standard library
 /// [`SocketAddr`], [`SocketAddrV4`], and [`SocketAddrV6`] types.
+#[derive(clone::Clone)]
```
Thomasdezeeuw, line 17: `Clone` is part of the prelude so we don't the import `clone`.

```suggestion
#[derive(Clone)]
```

src/sockaddr.rs
```diff
@@ -1,6 +1,6 @@
 use std::mem::{self, size_of, MaybeUninit};
 use std::net::{SocketAddr, SocketAddrV4, SocketAddrV6};
-use std::{fmt, io};
+use std::{clone, fmt, io};
```
Thomasdezeeuw, line 3:
```suggestion
use std::{fmt, io};
```

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @nathaniel-bennett.

PR #316, opened 2022-06-23
Initial basic cmsg support for unix

### LINKED ISSUE #313: Support ancillary data over `sendmsg`, `recvmsg`
Should `socket2` support `sendmsg`/`recvmsg` including control ancillary data to expose things like `IP_TTL` and `IPV6_PKTINFO`.

I [took a quick stab at it](https://github.com/brunowonka/socket2/commit/b228b0babe47ac9358117b3e9287669deb6ab06b) to figure out how it'd look like. I only did the writing side, I think there's a lot of design work and discussion here and I'm more than half-convinced this should be an `enum` instead of what I wrote.

I think it'd be interesting to basically provide some safe wrappers around `CMSG` and the many control data types. Unclear if it's in the interest of the authors of the crate to maintain all of those, however. Also I have 0 experience with Windows here so I didn't even try satisfying it with my prototype, and I'm honestly not even sure where to begin there.

Interested to hear the maintainer's thoughts. I'm willing to iterate on design and implementation here if there is interest.

Note: This is somewhat related to #86 from what I gathered in the discussion over there.

brunowonka:
Also added `set_recv_ttl` while I'm here to use in testing. The other `RECVxxx`-type socket options might make more sense with control message support now.

I went back and forth on the shape of the public API for this quite a bit, and I also don't have access to a Windows machine to check.

Fixes #313

### DIFF
--- a/src/cmsg.rs
+++ b/src/cmsg.rs
@@ -226,11 +226,13 @@ impl Cmsg {
     fn level_type_size(&self) -> (libc::c_int, libc::c_int, libc::c_uint) {
         match self {
             #[cfg(not(any(target_os = "solaris", target_os = "illumos")))]
-            Cmsg::IpTos(_) => (
-                libc::IPPROTO_IP,
-                libc::IP_TOS,
-                std::mem::size_of::<u8>() as libc::c_uint,
-            ),
+            Cmsg::IpTos(_) => {
+                #[cfg(not(target_os = "macos"))]
+                let len = std::mem::size_of::<u8>();
+                #[cfg(target_os = "macos")]
+                let len = std::mem::size_of::<i32>();
+                (libc::IPPROTO_IP, libc::IP_TOS, len as libc::c_uint)
+            }
             #[cfg(not(any(target_os = "fuchsia", target_os = "solaris", target_os = "illumos")))]
             Cmsg::Ipv6PktInfo { .. } => (
                 libc::IPPROTO_IPV6,
@@ -248,7 +250,15 @@ impl Cmsg {
         match self {
             #[cfg(not(any(target_os = "solaris", target_os = "illumos")))]
             Cmsg::IpTos(tos) => {
-                buffer[0] = *tos;
+                #[cfg(not(target_os = "macos"))]
+                {
+                    buffer[0] = *tos;
+                }
+                #[cfg(target_os = "macos")]
+                {
+                    let value = *tos as i32;
+                    buffer.copy_from_slice(&value.to_ne_bytes()[..])
+                }
             }
             #[cfg(not(any(target_os = "fuchsia", target_os = "solaris", target_os = "illumos")))]
             Cmsg::Ipv6PktInfo { addr, ifindex } => {
@@ -283,8 +293,17 @@ impl Cmsg {
         match (cmsg_level, cmsg_type) {
             #[cfg(not(any(target_os = "solaris", target_os = "illumos")))]
             (libc::IPPROTO_IP, libc::IP_TOS) => {
-                assert_eq!(bytes.len(), std::mem::size_of::<u8>(), "{:?}", bytes);
-                Cmsg::IpTos(bytes[0])
+                // Different systems encode received TOS as char or int.
+                match bytes {
+                    [b] => Cmsg::IpTos(*b),
+                    [a, b, c, d] => Cmsg::IpTos(i32::from_ne_bytes([*a, *b, *c, *d]) as u8),
+                    other => panic!("unexpected length for IP_TOS: {:?}", other),
+                }
+            }
+            #[cfg(any(target_os = "freebsd", target_os = "macos"))]
+            (libc::IPPROTO_IP, libc::IP_RECVTOS) => {
+                // Some systems use IP_RECVTOS on the receive path.
+                Self::from_raw(libc::IPPROTO_IP, libc::IP_TOS, bytes)
             }
             #[cfg(not(any(target_os = "fuchsia", target_os = "solaris", target_os = "illumos")))]
             (libc::IPPROTO_IPV6, libc::IPV6_PKTINFO) => {
--- a/src/cmsg.rs
+++ b/src/cmsg.rs
@@ -49,13 +49,14 @@ impl<B: AsRef<[u8]>> MsgHdrWalker<B> {
             // SAFETY: cmsghdr is a valid pointer given to us by `next_ptr`.
             let data = unsafe { libc::CMSG_DATA(cmsghdr) };
             let cmsghdr = unsafe { &*cmsghdr };
+            // SAFETY: Only copied values; need to grab the baseline for length.
+            let hdr_len = unsafe { libc::CMSG_LEN(0) } as usize;
             // SAFETY: data points to buffer and is controlled by control
             // message length.
             let data = unsafe {
                 std::slice::from_raw_parts(
                     data,
-                    (cmsghdr.cmsg_len as usize)
-                        .saturating_sub(std::mem::size_of::<libc::cmsghdr>()),
+                    (cmsghdr.cmsg_len as usize).saturating_sub(hdr_len),
                 )
             };
             (cmsghdr, data)
--- a/src/cmsg.rs
+++ b/src/cmsg.rs
@@ -1,0 +1,359 @@
+use std::convert::TryInto as _;
+use std::io::IoSlice;
+
+#[derive(Debug, Clone)]
+struct MsgHdrWalker<B> {
+    buffer: B,
+    position: Option<usize>,
+}
+
+impl<B: AsRef<[u8]>> MsgHdrWalker<B> {
+    fn next_ptr(&mut self) -> Option<*const libc::cmsghdr> {
+        // Build a msghdr so we can use the functionality in libc.
+        let mut msghdr: libc::msghdr = unsafe { std::mem::zeroed() };
+        let buffer = self.buffer.as_ref();
+        // SAFETY: We're giving msghdr a mutable pointer to comply with the C
+        // API. We'll only allow mutation of `cmsghdr`, however if `B` is
+        // AsMut<[u8]>.
+        msghdr.msg_control = buffer.as_ptr() as *mut _;
+        msghdr.msg_controllen = buffer.len().try_into().expect("buffer is too long");
+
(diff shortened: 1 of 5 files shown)


src/socket.rs
```diff
@@ -1415,6 +1472,31 @@ impl Socket {
...
+    /// incoming packets.
+    #[cfg(not(any(
+        target_os = "fuschia",
+        target_os = "illumos",
+        target_os = "netbsd",
+        target_os = "openbsd",
+        target_os = "redox",
+        target_os = "solaris",
+    )))]
+    pub fn set_recv_ttl(&self, recv_ttl: bool) -> io::Result<()> {
```
Thomasdezeeuw, line 1487: Can you move this to another pr, that would be quicker to merge, and please also add a getter, following the same pattern for all the other options.

src/socket.rs
```diff
@@ -660,6 +662,61 @@ impl Socket {
     ) -> io::Result<usize> {
         sys::send_to_vectored(self.as_raw(), bufs, addr, flags)
     }
+
+    /// Sends data on the socket accompanied by ancillary control message data.
+    #[cfg(unix)]
+    pub fn send_msg<B: AsRef<[u8]>>(
```
Thomasdezeeuw, line 668: Let's remove the generic parameter and use `&[u8]` directly instead.

src/socket.rs
```diff
@@ -660,6 +662,61 @@ impl Socket {
...
+        &self,
+        addr: Option<&SockAddr>,
```
Thomasdezeeuw, line 670: Let's split this code into `send_msg` (without an address) and `send_msg_to` (with an address), following the same pattern of the other send functions.

src/cmsg.rs
```diff
@@ -0,0 +1,365 @@
+use crate::sys;
+use std::borrow::Borrow;
+use std::convert::TryInto as _;
+use std::io::IoSlice;
+use std::iter::FromIterator;
+
+#[derive(Debug, Clone)]
+struct MsgHdrWalker<B> {
```
Thomasdezeeuw, line 8: Let's remove the generic parameter, we only need `&[u8]`, same for the other structures.

src/sys/unix.rs
```diff
@@ -137,6 +128,15 @@ pub(crate) use libc::{IPV6_ADD_MEMBERSHIP, IPV6_DROP_MEMBERSHIP};
...
 };
+#[cfg(not(any(
+    target_os = "fuschia",
+    target_os = "illumos",
+    target_os = "netbsd",
+    target_os = "openbsd",
+    target_os = "redox",
+    target_os = "solaris",
+)))]
+pub(crate) use libc::{IP_RECVTOS, IP_RECVTTL};
```
Thomasdezeeuw, line 139: Seems like the constant `IP_RECVTTL` is not available on various platform.

src/sys/unix.rs
```diff
@@ -845,6 +862,10 @@ fn sendmsg(
     // Safety: Same as above about `*const` -> `*mut`.
     msg.msg_iov = bufs.as_ptr() as *mut _;
     msg.msg_iovlen = min(bufs.len(), IovLen::MAX as usize) as IovLen;
+    // Safety: Same as above about `*const` -> `*mut`.
+    msg.msg_control = control_data.as_ptr() as *mut _;
+    msg.msg_controllen = control_data.len();
+    println!("LENGTH = {}", control_data.len());
```
Thomasdezeeuw, line 868: Remove.

src/sys/unix.rs
```diff
@@ -845,6 +862,10 @@ fn sendmsg(
     // Safety: Same as above about `*const` -> `*mut`.
     msg.msg_iov = bufs.as_ptr() as *mut _;
     msg.msg_iovlen = min(bufs.len(), IovLen::MAX as usize) as IovLen;
+    // Safety: Same as above about `*const` -> `*mut`.
+    msg.msg_control = control_data.as_ptr() as *mut _;
+    msg.msg_controllen = control_data.len();
```
Thomasdezeeuw, line 867: At least on macOS this is a `u32`.
```suggestion
    msg.msg_controllen = control_data.len() as _;
```

src/sys/unix.rs
```diff
@@ -785,8 +786,16 @@ fn recvmsg(
     msg.msg_namelen = msg_namelen;
     msg.msg_iov = bufs.as_mut_ptr().cast();
     msg.msg_iovlen = min(bufs.len(), IovLen::MAX as usize) as IovLen;
-    syscall!(recvmsg(fd, &mut msg, flags))
-        .map(|n| (n as usize, msg.msg_namelen, RecvFlags(msg.msg_flags)))
+    msg.msg_control = control_data.as_mut_ptr().cast();
+    msg.msg_controllen = control_data.len();
```
Thomasdezeeuw, line 790: At least on macOS this is a `u32`.
```suggestion
    msg.msg_controllen = control_data.len() as _;
```

src/sys/unix.rs
```diff
@@ -785,8 +786,16 @@ fn recvmsg(
...
+    syscall!(recvmsg(fd, &mut msg, flags)).map(|n| {
+        (
+            n as usize,
+            msg.msg_namelen,
+            msg.msg_controllen,
```
Thomasdezeeuw, line 795:
```suggestion
            msg.msg_controllen as libc::size_t,
```

Thomasdezeeuw (changes requested):
I'm afraid this doesn't work at all on a number of platform, please see the CI failure. I'll do another round of review once those are fixed.

src/socket.rs
brunowonka: I only introduced this because I wanted some control message to use in tests, I've moved to use `recv_tos` and `IP_TOS` instead so I don't need this.

src/sys/unix.rs
brunowonka: Reverted this.

src/cmsg.rs
brunowonka: Tentatively keeping this parameter so we can reuse some logic between `&[u8]` and `&mut [u8]`, but removed all the type parameters from the public API for now.

brunowonka:
Thank you for looking, I'm working through the other platform problems here and I'll push as soon as I have things working so you can look through.

tests/socket.rs
```diff
@@ -1265,3 +1265,60 @@ fn header_included() {
...
+
+#[test]
+#[cfg(all(
+    unix,
+    not(any(
+        target_os = "fuchsia",
+        target_os = "solaris",
+        target_os = "illumos",
+        target_os = "netbsd",
+        target_os = "redox",
```
Thomasdezeeuw, line 1277: This list should match the list in the code, which currently only include redox. If their is a reason for the exclusion please add a comment.

src/socket.rs
```diff
@@ -660,6 +662,76 @@ impl Socket {
...
+    }
+
+    /// Receives data on the socket accompanied by ancillary control message data.
+    #[cfg(all(unix, not(target_os = "redox")))]
+    pub fn recv_msg(
+        &self,
+        bufs: &mut [MaybeUninitSlice<'_>],
+        control_data: &mut [MaybeUninit<u8>],
+        flags: c_int,
+    ) -> io::Result<(usize, SockAddr, CmsgIter<'_>, RecvFlags)> {
```
Thomasdezeeuw, line 712: Can you match the other recv functions `(usize, CmsgIter<'_>, RecvFlags, SockAddr)`?

src/socket.rs
```diff
@@ -660,6 +662,76 @@ impl Socket {
...
+            addr.len(),
+            bufs,
+            cmsg.io_slice(),
+            flags,
+        )
+    }
+
+    /// Receives data on the socket accompanied by ancillary control message data.
+    #[cfg(all(unix, not(target_os = "redox")))]
+    pub fn recv_msg(
```
Thomasdezeeuw, line 707: Since this get an address in return it should be called `recv_msg_from`.

src/socket.rs
```diff
@@ -660,6 +662,76 @@ impl Socket {
...
+        &self,
+        bufs: &mut [MaybeUninitSlice<'_>],
+        control_data: &mut [MaybeUninit<u8>],
```
Thomasdezeeuw, line 710: Not sure about this type, would be nice if we could provide a nice wrapper around it.

Thomasdezeeuw (commented):
Not a complete review, but a start.

tests/socket.rs
brunowonka: Renamed the test and added a note. We're limited by enabling IP_TOS being generated here

darconeous:
Looking at [this failure](https://github.com/rust-lang/socket2/runs/7569907645?check_suite_focus=true):

```
Run cargo install cargo-hack
    Updating crates.io index
 Downloading crates ...
  Downloaded cargo-hack v0.5.15
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec[8](https://github.com/rust-lang/socket2/runs/7569907645?check_suite_focus=true#step:4:9)23/cargo-hack-0.5.15/Cargo.toml`

Caused by:
  failed to parse the `edition` key

Caused by:
  this version of Cargo is older than the `2021` edition, and only supports `2015` and `2018` editions.
Error: Process completed with exit code [10](https://github.com/rust-lang/socket2/runs/7569907645?check_suite_focus=true#step:4:11)1.
```

Looks unrelated to this change?

The [macOS failure](https://github.com/rust-lang/socket2/runs/7569907933?check_suite_focus=true) looks real though:

```
running 36 tests
...snip...
test send_from_recv_to_vectored ... FAILED
test send_recv_vectored ... FAILED
...snip...

failures:
error: test failed, to rerun pass '--test socket'

error: process didn't exit successfully: `/Users/runner/.rustup/toolchains/stable-x86_64-apple-darwin/bin/cargo test --manifest-path Cargo.toml --no-default-features` (exit status: 101)
---- send_from_recv_to_vectored stdout ----
thread 'send_from_recv_to_vectored' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }', tests/socket.rs:[57](https://github.com/rust-lang/socket2/runs/7569907933?check_suite_focus=true#step:5:58)4:10
```

Thomasdezeeuw:
Closing due to age.

PR #322, opened 2022-07-27
Use pre-built cargo-hack in CI

LingMan:
Use pre-built cargo-hack in CI
    
The authors of cargo-hack provide an Action to use in CI and avoid having to build from source.
Currently the MSRV CI job fails because latest cargo-hack requires 1.56 to build while the test is
run with 1.46.
Simply running cargo-hack only requies 1.26 though.
    
This should also speedup CI runs, since we don't have to build cargo-hack each time.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,7 +41,7 @@ jobs:
       run: rustup update ${{ matrix.rust }} --no-self-update && rustup default ${{ matrix.rust }}
       shell: bash
     - name: Install cargo-hack
-      run: cargo install cargo-hack
+      uses: taiki-e/install-action@cargo-hack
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
I've just started the v0.4.x branch so it's not actually used yet. So let's wait until we merge https://github.com/rust-lang/socket2/pull/320 and we actually start with the master/v0.4 split.

LingMan:
Alright, guess you'll add the v0.4.x branch to CI with #320 then. I've dropped the second commit and updated the PR description.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @LingMan.

PR #328, opened 2022-08-13
Rename SockAddr::init to try_init

Thomasdezeeuw:
To match the function signature, which returns a result. Also allows to
add an `init` function that never fails, if ever needed.

Updates #231.


chansuke (approved):
Looks good to me

Darksonn (approved):

PR #334, opened 2022-08-20
Expand SockAddr API

Thomasdezeeuw:
See the individual commits.

Fixes #333.

PR #335, opened 2022-08-20
Fix a number of Clippy warnings

Thomasdezeeuw:
Includes #334, so ignore the first four commits.


Darksonn (approved):
Since this includes #334, how do you plan to merge it? Seems like it would be best to remove those commits, then rebase the other PR when you merge this.

Issue #49, opened 2020-01-12
Support TCP Fast Open


zonyitoo:
## `TcpListener`

All of Linux, OS X and Windows are the same, call `setsockopt` with `TCP_FASTOPEN` option

```c
// For Linux, value is the queue length of pending packets
int opt = 5;
// For others, just a boolean value for enable and disable
int opt = 1;

// Call it before listen()
int ret = setsockopt(socket, IPPROTO_TCP, TCP_FASTOPEN, &opt, sizeof(opt));
```

## `TcpStream`

This is a bit more complicated than `TcpListener`, because we have to send `SYN` with the first data packet to make TFO works. APIs are completely different on different platforms:

### Linux

Before 4.11, Linux uses `MSG_FASTOPEN` flag for `sendto()`, doesn't need to call `connect()`

```c
// Send SYN with data in buf
ssize_t n = sendto(socket, buf, buf_length, MSG_FASTOPEN, saddr, saddr_len);
```

After 4.11, Linux provides a `TCP_FASTOPEN_CONNECT` option

```c
int enable = 1;

// Enable it before connect()
int ret = setsockopt(socket, IPPROTO_TCP, TCP_FASTOPEN_CONNECT, &enable, sizeof(enable));

// Call connect as usual
int ret = connect(socket, saddr, saddr_len);
```

### BSDs (except Darwin)

Uses `sendto()` as Linux without `MSG_FASTOPEN` flag

```c
// Set before sendto()
int ret = setsockopt(socket, IPPROTO_TCP, TCP_FASTOPEN, &opt, sizeof(opt));

// Call sendto() without flags
ssize_t n = sendto(socket, buf, buf_length, 0, saddr, saddr_len);
```

### Darwin (OS X)

Darwin supports TFO after OS X 10.11, iOS 9.0, tvOS 9.0 and watchOS 2.0.

It supports with a new syscall `connectx`.

```c
sa_endpoints_t endpoints;
memset(&endpoints, 0, sizeof(endpoints));
endpoints.sae_dstaddr = &saddr;
endpoints.sae_dstaddrlen = saddr_len;

int ret = connectx(socket,
                   &endpoints,
                   SAE_ASSOCID_ANY,
                   CONNECT_DATA_IDEMPOTENT | CONNECT_RESUME_ON_READ_WRITE,
                   NULL,
                   0,
                   NULL,
                   NULL);
```

SYN will be sent with the first `write` or `send`.

### Windows

[...]

PR #336, opened 2022-08-24
TCP_FASTOPEN option for Socket

zonyitoo:
- Supported platforms: Linux-like, FreeBSD, macOS, Windows
- ref #49

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -1457,6 +1457,81 @@ impl Socket {
                 .map(|recv_tos| recv_tos > 0)
         }
     }
+
+    /// Set `TCP_FASTOPEN` option for this socket.
+    ///
+    /// ## Windows
+    ///
+    /// Windows supports TCP Fast Open since Windows 10.
+    ///
+    /// <https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-tcp-socket-options>
+    ///
+    /// `value` is a boolean with only `0` and `1`.
+    ///
+    /// ## Linux
+    ///
+    /// Linux supports TCP Fast Open since 3.7.
+    ///
+    /// <https://lwn.net/Articles/508865/>
+    ///
+    /// The option `value`, `qlen`, specifies this server's limit on the size of the queue of TFO requests that have
+    /// not yet completed the three-way handshake (see the remarks on prevention of resource-exhaustion attacks above).
+    ///
+    /// It was recommended to be `5` in this document.
+    ///
+    /// ## macOS
+    ///
+    /// `value` is a boolean with only `0` and `1`.
+    ///
+    /// ## FreeBSD
+    ///
+    /// FreeBSD supports TCP Fast Open since 12.0.
+    ///
+    /// Example program: <https://people.freebsd.org/~pkelsey/tfo-tools/tfo-srv.c>
+    ///
+    /// `value` is a boolean with only `0` and `1`.
+    #[cfg(any(
+        target_os = "linux",
+        target_os = "android",
+        target_os = "freebsd",
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "watchos",
+        target_os = "tvos",
+        target_os = "windows"
+    ))]
+    pub fn set_tcp_fastopen(&self, value: u32) -> io::Result<()> {
+        unsafe {
+            setsockopt::<c_int>(
+                self.as_raw(),
+                sys::IPPROTO_TCP,
+                sys::TCP_FASTOPEN,
+                value as c_int,
+            )
+        }
+    }
+
+    /// Get the value of `TCP_FASTOPEN` option for this socket.
+    ///
+    /// For more information about this option, see [`set_tcp_fastopen`].
+    ///
+    /// [`set_tcp_fastopen`]: Socket::set_tcp_fastopen
+    #[cfg(any(
+        target_os = "linux",
+        target_os = "android",
+        target_os = "freebsd",
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "watchos",
+        target_os = "tvos",
+        target_os = "windows"
+    ))]
+    pub fn tcp_fastopen(&self) -> io::Result<u32> {
+        unsafe {
+            getsockopt::<c_int>(self.as_raw(), sys::IPPROTO_TCP, sys::TCP_FASTOPEN)
+                .map(|c| c as u32)
+        }
+    }
 }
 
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw (commented):
Can you add a smoke test for this?

If I call this with the value `5`, does this work on non-Linux? I.e. enable it, or would it error and would users have to do `if linux { 5 } else { 1 }`? Depending on the answer can you also add a test for this?

zonyitoo:
```
thread 'tcp_fastopen' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }', tests/socket.rs:1302:32
```

We have the answer, `value` must be `0` and `1` excepts Linux.

zonyitoo:
```
thread 'tcp_fastopen' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" }', tests/socket.rs:1306:36
```

Test failed on FreeBSD CI.

https://github.com/freebsd/freebsd-src/blob/e7d02be19d40063783d6b8f1ff2bc4c7170fd434/sys/netinet/tcp_usrreq.c#L2458-L2507

It seems that TFO is not enabled on CI's OS.

zonyitoo:
@Thomasdezeeuw Should I remove FreeBSD support?

Thomasdezeeuw:
Can you change the method so that `0` means disable, and `>= 1` means enable, mapping anything bigger than 1 to 1 on non-Linux systems? And update the docs accordingly.

For FreeBSD can you try add a new step to [`.cirrus.yml`](https://github.com/rust-lang/socket2/blob/master/.cirrus.yml)? With the following:

```yml
  enable_tcp_fastopen:
    - sysctl net.inet.tcp.fastopen.client_enabled=1
    - sysctl net.inet.tcp.fastopen.server_enabled=1
```

After `build_script` should work. This should enable it on FreeBSD.

src/socket.rs
```diff
@@ -1500,7 +1500,11 @@ impl Socket {
         target_os = "tvos",
         target_os = "windows"
     ))]
-    pub fn set_tcp_fastopen(&self, value: u32) -> io::Result<()> {
+    pub fn set_tcp_fastopen(&self, mut value: u32) -> io::Result<()> {
+        if !cfg!(any(target_os = "linux", target_os = "android")) && value > 1 {
```
Thomasdezeeuw, line 1504: We don't use `cfg!`
```suggestion
        #[cfg(not(any(target_os = "linux", target_os = "android")))]
        if value > 1 {
```

If that doesn't work it will need block around it.

tests/socket.rs
```diff
@@ -1281,3 +1281,29 @@ fn header_included() {
...
+    let bsaddr = SockAddr::from(baddr);
+    socket.bind(&bsaddr).unwrap();
+    socket.listen(128).unwrap();
+    socket.set_tcp_fastopen(5).unwrap();
+
+    if cfg!(any(target_os = "linux", target_os = "android")) {
+        assert_eq!(socket.tcp_fastopen().unwrap(), 5);
+    } else {
+        assert_eq!(socket.tcp_fastopen().unwrap(), 1);
+    }
```
Thomasdezeeuw, line 1308:
```suggestion
    #[cfg(not(target_os = "linux", target_os = "android"))]
    let expected = 5;
    #[cfg(not(any(target_os = "linux", target_os = "android")))]
    let expected = 1;
    assert_eq!(socket.tcp_fastopen().unwrap(), expected);
```

zonyitoo:
Still not working on FreeBSD's CI test.

Thomasdezeeuw:
@zonyitoo I'll work on FreeBSD

Thomasdezeeuw:
@zonyitoo the `sysctl` names I gave you are incorrect. These should fix them:

```diff
-    - sysctl net.inet.tcp.fastopen.client_enabled=1
-    - sysctl net.inet.tcp.fastopen.server_enabled=1
+    - sudo sysctl net.inet.tcp.fastopen.client_enable=1
+    - sudo sysctl net.inet.tcp.fastopen.server_enable=1
```

However that doesn't seem to fix the problem (see my next comment).

Thomasdezeeuw:
@asomers maybe you can help here. I got this to work on FreeBSD (the `sysctl` names are incorrect). But when retrieving the value of `TCP_FASTOPEN` isn't 1, but `2147483648`, is this expected? I think the value comes from https://github.com/freebsd/freebsd-src/blob/e7d02be19d40063783d6b8f1ff2bc4c7170fd434/sys/netinet/tcp_usrreq.c#L2693-L2697, but I could be wrong here.

asomers:
@Thomasdezeeuw this socket option is a flag, so you should only interpret it as either zero or non-zero.  TCP_NODELAY is similar, and I see that socket2 interprets it correctly.  I suggest you change the `Socket::tcp_fastopen` signature to match `Socket::nodelay`.

Thomasdezeeuw:
> @Thomasdezeeuw this socket option is a flag, so you should only interpret it as either zero or non-zero. TCP_NODELAY is similar, and I see that socket2 interprets it correctly. I suggest you change the `Socket::tcp_fastopen` signature to match `Socket::nodelay`.

Thanks. That makes the API a little awkward as the value on Linux is the accept queue length, so we'll need to rethink the API.

.cirrus.yml
```diff
@@ -17,8 +17,8 @@ task:
     - cargo build
     - cargo build --no-default-features
   enable_tcp_fastopen:
-    - sysctl net.inet.tcp.fastopen.client_enabled=1
-    - sysctl net.inet.tcp.fastopen.server_enabled=1
+    - sudo sysctl net.inet.tcp.fastopen.client_enable=1
```
asomers, line 20: Is sudo really necessary?  Cirrus runs normally start as root.

redactedontop:
Updates?

zonyitoo:
Just resolved conflicts with the master branch. Please review again @Thomasdezeeuw

zonyitoo:
```
thread 'tcp_fastopen' panicked at tests/socket.rs:1639:32:
called `Result::unwrap()` on an `Err` value: Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" }
```

It seems that `net.inet.tcp.fastopen.client_enable` and `net.inet.tcp.fastopen.server_enable` were still not set properly. @asomers

Thomasdezeeuw:
Closing this due to inactivity.

Issue #308, opened 2022-05-10
Release v0.4.5 did a break change


bachue:
The PR "Replace `winapi` dependency with `windows-sys`"  (https://github.com/rust-lang/socket2/pull/303) breaks public apis, cause [`dns-lookup`](https://github.com/keeperofdakeys/dns-lookup) are failed to compile on windows https://github.com/keeperofdakeys/dns-lookup/issues/22. `dns-lookup` calls `getnameinfo` from `winapi`, but now `socket2` exports `SockAddr` from `windows-sys`, which are incompatible with `SockAddr` from `winapi`.

Thomasdezeeuw:
I've yanked v0.4.5 for now.

bachue:
@Thomasdezeeuw  👍

sn99:
@Thomasdezeeuw I am working on shifting `dns-lookup` from `winapi` to `window_sys`, I am using your master branch instead of crates.io and it seems to be working, I will make a pull in `dns-lookup` soon.

Thomasdezeeuw:
> @Thomasdezeeuw I am working on shifting `dns-lookup` from `winapi` to `window_sys`, I am using your master branch instead of crates.io and it seems to be working, I will make a pull in `dns-lookup` soon.

Sounds good, but unfortunately we still can't make a breaking change in a v0.4.x release, so the switch to window_sys will have to go into v0.5.

sn99:
> > @Thomasdezeeuw I am working on shifting `dns-lookup` from `winapi` to `window_sys`, I am using your master branch instead of crates.io and it seems to be working, I will make a pull in `dns-lookup` soon.
> 
> Sounds good, but unfortunately we still can't make a breaking change in a v0.4.x release, so the switch to window_sys will have to go into v0.5.

Oh yeah, no problem, just wanted to inform you about the change in dns-lookup

Thomasdezeeuw:
[...]

PR #340, opened 2022-08-26
Revert back to winapi dependency for v0.4

Thomasdezeeuw:
As noted in <https://github.com/rust-lang/socket2/issues/308>, switching
from winapi to windows-sys was a breaking change as we expose some of
the winapi types that are different in windows-sys.

This reverts the following commits:
 * "Relax windows-sys dependency version" 81056b9c8b579c60a5760e1971032b2c6e3d6488.
 * "Support Socket::(set_)recv_tos on Windows" 242e087168ab511b08e331ff004a1d8e789e2e95.
 * "Replace winapi with windows-sys" aeb6c2dd3befb5e43b2d4d05f362fe6ca61653e3.

We'll make this switch in v0.5 instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,14 +35,8 @@ features = ["all"]
 [target."cfg(unix)".dependencies]
 libc = "0.2.124"
 
-[target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.36"
-features = [
-  "Win32_Foundation",
-  "Win32_Networking_WinSock",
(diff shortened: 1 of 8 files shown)


Darksonn (approved):
LGTM, but wasn't this already reverted?

Thomasdezeeuw:
> LGTM, but wasn't this already reverted?

No, we just yanked the v0.4.5 release.

PR #343, opened 2022-08-30
V0.4.6 openbsd

niklasha:
0.4.6 broke on OpenBSD.  Besides I think the tests were broken even earlier.  I made some fixes to the conditional compilation configuration in the areas:
* per-socket keepalive parameters
* Ipv4 source specific multicast
* RECVTOS
* ipv6 only
Tested on a system derived from OpenBSD 7.1, the latest published OpenBSD version.


Thomasdezeeuw:
Can you target the main branch and we'll back port the changes. I don't want to lose commits in the main branch.

niklasha:
> Can you target the main branch and we'll back port the changes. I don't want to lose commits in the main branch.

Done so, will close this PR.

PR #344, opened 2022-08-31
V0.5 alpha openbsd

niklasha:
For OpenBSD, current master is broken.  I made some fixes to the conditional compilation configuration in the following areas:

per-socket keepalive parameters
Ipv4 source specific multicast
RECVTOS
ipv6 only
Tested on a system derived from OpenBSD 7.1, the latest published OpenBSD version.


Thomasdezeeuw:
Thanks @niklasha

PR #349, opened 2022-09-08
support IPPROTO_MPTCP

Martichou:
Simply add the support for MPTCP protocol by adding and exposing the MPTCP const Protocol.

MPTCP allow end user to open multiple path (TCP) to the destination server using multiple interface. For example, one can use the LTE connection of his phone as well as a Wifi connection. Or Ethernet and Wifi.

Setting this Protocol will fail if the host does not support MPTCP, thus the fallback is to open as TCP like one would normally do.

For reference: `sysctl net.mptcp.enabled=1`

I'm not sure about adding test as this imply that the host must support MPTCP, open for feedback on this :)

Signed-off-by: Martin Andre <martin.andre@tessares.net>

### DIFF
--- a/tests/socket.rs
+++ b/tests/socket.rs
@@ -106,6 +106,8 @@ fn protocol_fmt_debug() {
         (Protocol::ICMPV6, "IPPROTO_ICMPV6"),
         (Protocol::TCP, "IPPROTO_TCP"),
         (Protocol::UDP, "IPPROTO_UDP"),
+        #[cfg(target_os = "linux")]
+        (Protocol::MPTCP, "IPPROTO_MPTCP"),
         (500.into(), "500"),
     ];
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -302,6 +302,10 @@ impl Protocol {
 
     /// Protocol corresponding to `UDP`.
     pub const UDP: Protocol = Protocol(sys::IPPROTO_UDP);
+
+    #[cfg(target_os = "linux")]
+    /// Protocol corresponding to `MPTCP`.
+    pub const MPTCP: Protocol = Protocol(sys::IPPROTO_MPTCP);
 }
 
 impl From<c_int> for Protocol {
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -63,6 +63,8 @@ pub(crate) use libc::SOCK_RAW;
 pub(crate) use libc::SOCK_SEQPACKET;
 pub(crate) use libc::{SOCK_DGRAM, SOCK_STREAM};
 // Used in `Protocol`.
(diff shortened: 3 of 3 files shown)


tests/socket.rs
```diff
@@ -101,20 +101,23 @@ fn type_fmt_debug() {
...
+    let mut tests = vec![
         (Protocol::ICMPV4, "IPPROTO_ICMP"),
         (Protocol::ICMPV6, "IPPROTO_ICMPV6"),
         (Protocol::TCP, "IPPROTO_TCP"),
         (Protocol::UDP, "IPPROTO_UDP"),
         (500.into(), "500"),
     ];
 
+    #[cfg(target_os = "linux")]
+    tests.push((Protocol::MPTCP, "IPPROTO_MPTCP"));
```
Thomasdezeeuw, line 113: I don't think you have to make `tests` a vector. I think you can do the following
```rust
        #[cfg(target_os = "linux")]
        (Protocol::MPTCP, "IPPROTO_MPTCP"),
```

Thomasdezeeuw (approved):
Do know if any other Linux like OS supports this? For example Android or Fuschia?

tests/socket.rs
Martichou: You're rigth! I've updated the commit :)

Martichou:
> Do know if any other Linux like OS supports this? For example Android or Fuschia?

It all depends on the Kernel, at the moment only Linux Kernel >=5.6 have MPTCP.
I can see in the Rust's `libc` that `l4re` seems to share the same export as `linux` (see https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/mod.rs#L1818), thus they also share the `libc::IPPROTO_MPTCP` const, but I know nothing about `l4re` so I'm not sure.

As Android is based on the Linux Kernel it should be compatible, but the Rust libc team did not define the `IPPROTO_MPTCP` for Android. I think that support for others OS can be added later if found compatible.

Thomasdezeeuw:
> > Do know if any other Linux like OS supports this? For example Android or Fuschia?
> 
> It all depends on the Kernel, at the moment only Linux Kernel >=5.6 have MPTCP. I can see in the Rust's `libc` that `l4re` seems to share the same export as `linux` (see https://github.com/rust-lang/libc/blob/master/src/unix/linux_like/mod.rs#L1818), thus they also share the `libc::IPPROTO_MPTCP` const, but I know nothing about `l4re` so I'm not sure.

We don't support l4re.

> As Android is based on the Linux Kernel it should be compatible, but the Rust libc team did not define the `IPPROTO_MPTCP` for Android. I think that support for others OS can be added later if found compatible.

We can add Android later.

Thomasdezeeuw:
Thanks @Martichou.

PR #352, opened 2022-09-21
Adding DCCP support

onestay:
This pull request adds support for the DCCP protocol. 
This included adding support for the DCCP Protocol and Type as well as  adding a bunch of socket options. Not all socket options as listed on https://www.kernel.org/doc/html/latest/networking/dccp.html are implemented. `DCCP_SOCKOPT_QPOLICY_ID` is not added because I could not find the constants `DCCPQ_POLICY_PRIO`, `DCCPQ_POLICY_SIMPLE` in the libc crate.

 `DCCP_SOCKOPT_CCID_RX_INFO` and `DCCP_SOCKOPT_CCID_TX_INFO` are also not added since they are CCID3 exclusive. 

Setting different RX and TX CCIDs is also not supported.  This is, in theory, allowed, however in practice very rarely used and not well understood.

I also wrote a simple test, testing _most_ socket options, as well as creating a server and client and sending a message between the two.


src/lib.rs
```diff
@@ -257,6 +257,12 @@ impl Type {
     /// Used for protocols such as UDP.
     pub const DGRAM: Type = Type(sys::SOCK_DGRAM);
 
+    #[cfg(target_os = "linux")]
```
Thomasdezeeuw, line 260: Can you put the `cfg` after the docs. This is a convention in Rust code.

Thomasdezeeuw, line 260: Can you also add a requirement on the `all` feature. This is used by socket to expose features that aren't available on all platforms, letting users know without workaround their application might not be cross-platform. See for example `SEQPACKET` just below (the `cfg_attr(docsrs, ..)` will also be needed.

Thomasdezeeuw: The two comment above also apply to all the other constants/functions.

src/sys/unix.rs
```diff
@@ -1506,6 +1511,22 @@ impl crate::Socket {
         }
     }
 
+    /// Returns list of CCIDs supported by the endpoint
+    #[cfg(target_os = "linux")]
+    pub fn dccp_available_ccids(&self) -> io::Result<Vec<u8>> {
+        let mut buf: [MaybeUninit<u8>; 10] = unsafe { MaybeUninit::uninit().assume_init() };
```
Thomasdezeeuw, line 1517: Could you add a comment and a maybe a link to some docs on why 10?

src/sys/unix.rs
```diff
@@ -1506,6 +1511,22 @@ impl crate::Socket {
...
+        let mut len = buf.len() as libc::socklen_t;
+        syscall!(getsockopt(
+            self.as_raw(),
+            libc::SOL_DCCP,
+            libc::DCCP_SOCKOPT_AVAILABLE_CCIDS,
+            buf.as_mut_ptr().cast(),
+            &mut len,
+        ))?;
+        let buf = &buf[..len as usize - 1];
+        Ok(unsafe { &*(buf as *const [_] as *const [u8]) }.into())
```
Thomasdezeeuw, line 1527: Can we do better than allocating 10 bytes? Maybe a small vec like type?

Thomasdezeeuw (commented):
Initial review comments.

Note that you don't need to add all options in one go. It's perfectly fine to add them in follow up prs (or not at all ;) )

src/lib.rs
onestay: So all the functions and constants related to dccp should have this?
```rust
    #[cfg(all(feature = "all", target_os = "linux"))]
    #[cfg_attr(docsrs, doc(cfg(feature = "all", target_os = "linux")))]
```

Thomasdezeeuw: Yes

src/sys/unix.rs
onestay: I'm not quiet sure what to do here. I've changed the buf to length 5 instead of 10. The code of this function was basically copied from the device function.
https://github.com/rust-lang/socket2/blob/4c436818cddec880c26c4e79aca7106ba88827ff/src/sys/unix.rs#L1488-L1507

Thomasdezeeuw (commented):
Can you follow our code conventions? That is
* Start setters with `set_`, e.g. `set_dccp_service`.
* Setters are before getters in the code.
* Add the socket option used to the function docs and maybe add a quick 1 sentence description. It's ok if this comes from the original docs.
* In the getter functions refer to the setter docs.

For example see https://github.com/rust-lang/socket2/blob/4c436818cddec880c26c4e79aca7106ba88827ff/src/sys/unix.rs#L1906-L1968

Thomasdezeeuw:
Sorry @onestay I missed this one. I took a quick glance and it looks pretty good. Can you rebase this on master and then I'll do a final review pass.

Thomasdezeeuw:
Thanks @onestay, this was merged as a6d6587d4ab4c22e0216172ca98b004b905be321.

PR #356, opened 2022-10-14
support IPPROTO_SCTP

### LINKED ISSUE #202: Support basic SCTP over plain/legacy sockets
Using just the basic socket APIs like connect/bind/send/etc it is possible to interact with SCTP sockets. While you don't get the full SCTP feature-set when using them, nevertheless they do work, so... thoughts?

nathaniel-bennett:
Adds the bare minimum needed to support SCTP sockets (IPPROTO_SCTP) as a protocol option for Linux and FreeBSD. Note that this does not include socket options specific to SCTP, as the constants and structs required have not been added to the libc crate.

Resolves #202

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -307,13 +307,7 @@ impl Protocol {
     /// Protocol corresponding to `MPTCP`.
     pub const MPTCP: Protocol = Protocol(sys::IPPROTO_MPTCP);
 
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @nathaniel-bennett

PR #357, opened 2022-11-15
Fix Socket example

pabigot:
Bind an IPv6 address to an IPv6 socket, and only do it once.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -55,12 +55,11 @@ use crate::{MaybeUninitSlice, RecvFlags};
 /// use std::net::{SocketAddr, TcpListener};
 /// use socket2::{Socket, Domain, Type};
 ///
-/// // create a TCP listener bound to two addresses
-/// let socket = Socket::new(Domain::IPV4, Type::STREAM, None)?;
+/// // create a TCP listener
+/// let socket = Socket::new(Domain::IPV6, Type::STREAM, None)?;
 ///
 /// let address: SocketAddr = "[::1]:12345".parse().unwrap();
 /// let address = address.into();
-/// socket.bind(&address)?;
 /// socket.bind(&address)?;
 /// socket.listen(128)?;
 ///


src/socket.rs
```diff
@@ -56,12 +56,11 @@ use crate::{MaybeUninitSlice, RecvFlags};
 /// use socket2::{Socket, Domain, Type};
 ///
 /// // create a TCP listener bound to two addresses
-/// let socket = Socket::new(Domain::IPV4, Type::STREAM, None)?;
+/// let socket = Socket::new(Domain::IPV6, Type::STREAM, None)?;
 ///
 /// let address: SocketAddr = "[::1]:12345".parse().unwrap();
 /// let address = address.into();
 /// socket.bind(&address)?;
-/// socket.bind(&address)?;
```
JohnTitor, line 64: I think this line shouldn't be removed as this example is for creating a TCP listener bound to two addresses, as the comment above says.

Thomasdezeeuw:
@pabigot what is broken about the example? Socket are able to be bound to two addresses.

pabigot:
> @pabigot what is broken about the example? Socket are able to be bound to two addresses.

@Thomasdezeeuw it doesn't work.  Try it in the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=821b177f130ce978b7b4e3854d98087d).

First, you get:

    Error: Os { code: 97, kind: Uncategorized, message: "Address family not supported by protocol" }

because it's trying to bind an IPv6 address to an IPv4 socket.  After fixing that, the playground doesn't work:

    Error: Os { code: 99, kind: AddrNotAvailable, message: "Cannot assign requested address" }

but if you move the example to a real system you get:
```
first bind
second bind
Error: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
```

@JohnTitor While the comment says it's an example for binding a socket to multiple addresses, [bind(2)](https://man7.org/linux/man-pages/man2/bind.2.html) says you get `EINVAL` if the socket is already bound to the address---which the example demonstrates.

Something as complex as binding two addresses to the same socket, if it can be done, seems inappropriate as the entry-point example for how to use Socket.  I've updated the comment; if you want this solved another way, let me know.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @pabigot, this makes the example less confusing at the very least.

PR #360, opened 2022-12-01
Add support for SO_ORIGINAL_DST and IP6T_SO_ORIGINAL_DST.

PiotrSikora:
Those values contain the original destination IPv4/IPv6 address
of the connection redirected using iptables REDIRECT or TPROXY.

Signed-off-by: Piotr Sikora <piotr@aviatrix.com>

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -1793,10 +1793,16 @@ impl crate::Socket {
     ///
     /// This value contains the original destination IPv4 address of the connection
     /// redirected using `iptables` `REDIRECT` or `TPROXY`.
-    #[cfg(all(feature = "all", any(target_os = "android", target_os = "linux")))]
+    #[cfg(all(
+        feature = "all",
+        any(target_os = "android", target_os = "fuchsia", target_os = "linux")
+    ))]
     #[cfg_attr(
         docsrs,
-        doc(cfg(all(feature = "all", any(target_os = "android", target_os = "linux"))))
+        doc(cfg(all(
+            feature = "all",
+            any(target_os = "android", target_os = "fuchsia", target_os = "linux")
+        )))
     )]
     pub fn original_dst(&self) -> io::Result<SockAddr> {
         // Safety: `getsockopt` initialises the `SockAddr` for us.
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
Could also add a test, even if it's ignored by default.

src/sys/unix.rs
```diff
@@ -1789,6 +1789,68 @@ impl crate::Socket {
...
+                    libc::SOL_IP,
+                    libc::SO_ORIGINAL_DST,
+                    storage.cast(),
+                    len
+                ))
+            })
+        }
+        .map_or_else(
+            |e| match e.raw_os_error() {
+                Some(libc::ENOENT) => Ok(None),
```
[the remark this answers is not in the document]
PiotrSikora: Done.

PiotrSikora:
> Could also add a test, even if it's ignored by default.

Done.

src/sys/unix.rs
```diff
@@ -1789,6 +1789,62 @@ impl crate::Socket {
         }
     }
 
+    /// Get the value for the `SO_ORIGINAL_DST` option on this socket.
+    ///
+    /// This value contains the original destination IPv4 address of the connection
+    /// redirected using `iptables` `REDIRECT` or `TPROXY`.
+    #[cfg(all(
+        feature = "all",
+        any(target_os = "android", target_os = "fuchsia", target_os = "linux")
```
PiotrSikora, line 1798: Note: I've added `fuchsia` based on presence of `libc::SO_ORIGINAL_DST` in Rust's libc for Fuchsia, but I have no way to verify that it actually works on that platform. Let me know if you want me to remove it.

tests/socket.rs
```diff
@@ -1297,3 +1297,39 @@ fn header_included() {
...
+
+#[test]
+#[cfg(all(
+    feature = "all",
+    any(target_os = "android", target_os = "fuchsia", target_os = "linux")
+))]
+fn original_dst() {
+    let socket = Socket::new(Domain::IPV4, Type::STREAM, None).unwrap();
+    match socket.original_dst() {
+        Ok(_) => panic!("original_dst on non-redirected socket should fail"),
```
Thomasdezeeuw, line 1309: How difficult would it be to add a test that successfully retrieved the address?

Thomasdezeeuw (approved):

tests/socket.rs
PiotrSikora: You'd need to configure `iptables` redirect in the operating system, which seems invasive for a test of the bindings.

Thomasdezeeuw:
Thanks @PiotrSikora.

PR #364, opened 2022-12-19
Add support for IPV6_TCLASS and IPV6_RECVTCLASS

### LINKED ISSUE #363: Support IPV6_TCLASS
`IPV6_TCLASS` is `IP_TOS` in IPv6. It is already defined in `libc` ([ref](https://docs.rs/libc/0.2.138/libc/constant.IPV6_TCLASS.html)).

A simple solution would be `pub(crate) use libc::IPV6_TCLASS;`. But I'm not sure how to correctly configure the cfg attribute.

duskmoon314:
A simple implementation of the IPV6_TCLASS and IPV6_RECVTCLASS modeled after the existing IP_TOS and IP_RECVTOS implementations.

close #363


src/socket.rs
```diff
@@ -1668,6 +1668,98 @@ impl Socket {
...
         }
     }
+
+    /// Set the value of the `IPV6_TCLASS` option for this socket.
+    ///
+    /// Specifies the traffic class field that is used in every packets
+    /// sent from this socket.
+    ///
+    /// NOTE: <https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options>
+    /// does not list this option, but it is supported in `windows-sys`.
```
Thomasdezeeuw, line 1678: Don't need to mention this.
```suggestion
```

src/socket.rs
```diff
@@ -1668,6 +1668,98 @@ impl Socket {
...
+        }
+    }
+
+    /// Get the value of the `IPV6_TCLASS` option for this socket.
+    ///
+    /// For more information about this option, see [`set_tclass_v6`].
+    ///
+    /// NOTE: <https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options>
+    /// does not list this option, but it is supported in `windows-sys`.
+    ///
```
Thomasdezeeuw, line 1702:
```suggestion
```

src/socket.rs
```diff
@@ -1668,6 +1668,98 @@ impl Socket {
...
+        target_os = "dragonfly",
+        target_os = "fuchsia",
+        target_os = "illumos",
+        target_os = "netbsd",
+        target_os = "openbsd",
+        target_os = "redox",
+        target_os = "solaris",
+    )))]
+    pub fn set_recv_tclass_v6(&self, recv_tclass: bool) -> io::Result<()> {
+        let recv_tclass = i32::from(recv_tclass);
```
Thomasdezeeuw, line 1732: Not needed, `bool as c_int` works.
```suggestion
```

Thomasdezeeuw (commented):
Can you switch the ordering of the functions, first the getter than the setter.

duskmoon314:
Done

Should I remove `let recv_tos = i32::from(recv_tos);` too?

Thomasdezeeuw:
> Should I remove `let recv_tos = i32::from(recv_tos);` too?

Sure 👍

Thomasdezeeuw (approved):

Thomasdezeeuw:
@duskmoon314 can you fix the failing CI for Windows? See https://github.com/rust-lang/socket2/actions/runs/3740167945/jobs/6348219101

duskmoon314:
Sure. I am looking for MS documentation on this setting

duskmoon314:
It seems that Microsoft doesn't allow modifying the traffic class of IPv6 win-socket as in Unix-like OS.

For IPv4, using `IP_TOS` is allowed and may correspond to history. And the doc notes that QoS API is recommended:

> Do not use. Type of Service (TOS) settings should only be set using the Quality of Service API. See [Differentiated Services](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/qos/differentiated-services) in the Quality of Service section of the Platform SDK for more information.
> 
> https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options

And for IPv6, no `IPV6_TCLASS` is mentioned in the doc. So this is probably why the test fails since I assume it is usable and just not be mentioned.

I am considering adding `target_os = windows` to the cfg `not` attribute. Any suggestions?

Thomasdezeeuw:
> I am considering adding `target_os = windows` to the cfg `not` attribute. Any suggestions?

Yes, that's the way to go I think. However if it's not available on Windows it need to be behind the `all` feature and we'll likely want to move it to `sys/unix.rs`. For an example see https://github.com/rust-lang/socket2/blob/fa67e04d6fdb3dca0075db2a88acbffe39980eee/src/sys/unix.rs#L1048-L1076

duskmoon314:
Sure. I will modify the code ASAP.

src/sys/unix.rs
```diff
@@ -76,6 +76,30 @@ pub(crate) use libc::{
...
+        target_os = "dragonfly",
+        target_os = "freebsd",
+        target_os = "fuchsia",
+        target_os = "linux",
+        target_os = "macos",
+        target_os = "netbsd",
+        target_os = "openbsd"
+    )
+))]
+pub(crate) use libc::IPV6_TCLASS;
```
Thomasdezeeuw, line 102: No longer needed.

```suggestion
```

src/sys/unix.rs
```diff
@@ -2057,6 +2081,84 @@ impl crate::Socket {
...
+                target_os = "linux",
+                target_os = "macos",
+                target_os = "netbsd",
+                target_os = "openbsd"
+            )
+        )))
+    )]
+    pub fn tclass_v6(&self) -> io::Result<u32> {
+        unsafe {
+            getsockopt::<c_int>(self.as_raw(), IPPROTO_IPV6, IPV6_TCLASS)
```
Thomasdezeeuw, line 2121:
```suggestion
            getsockopt::<c_int>(self.as_raw(), IPPROTO_IPV6, libc::IPV6_TCLASS)
```

src/sys/unix.rs
```diff
@@ -2057,6 +2081,84 @@ impl crate::Socket {
...
+                target_os = "fuchsia",
+                target_os = "linux",
+                target_os = "macos",
+                target_os = "netbsd",
+                target_os = "openbsd"
+            )
+        )))
+    )]
+    pub fn set_tclass_v6(&self, tclass: u32) -> io::Result<()> {
+        unsafe { setsockopt(self.as_raw(), IPPROTO_IPV6, IPV6_TCLASS, tclass as c_int) }
```
Thomasdezeeuw, line 2160:
```suggestion
        unsafe { setsockopt(self.as_raw(), IPPROTO_IPV6, libc::IPV6_TCLASS, tclass as c_int) }
```

Thomasdezeeuw (approved):

Thomasdezeeuw:
I'll squash the commit I've added.

Thomasdezeeuw:
Thanks @duskmoon314!

PR #367, opened 2022-12-21
Use std `spare_capacity_mut`

chansuke:
Use [Vec::spare_capacity_mut](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.spare_capacity_mut).


tests/socket.rs
```diff
@@ -942,15 +941,6 @@ fn sendfile() {
         target_vendor = "apple",
     )
 ))]
```
Thomasdezeeuw, line 943: Can you also remove the `cfg`

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @chansuke.

PR #369, opened 2022-12-22
Allow compilation on haiku beta 4

superboum:
I am running haiku beta 4 with rust installed from their package manager.
When trying to compile the matrix weechat rust plugin, the build failed while compiling the socket2 dependency.
After some investigation, it seems that haiku is lacking `libc::IPV6_RECVTCLASS` and ` libc::IP_RECVTOS`.
In this PR, I disable these 2 features for Haiku in the socket2 code, after that I was able to compile the library successfully.
However I did not test this code yet.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -1442,6 +1442,7 @@ impl Socket {
         target_os = "openbsd",
         target_os = "redox",
         target_os = "solaris",
+        target_os = "haiku",
     )))]
     pub fn set_recv_tos(&self, recv_tos: bool) -> io::Result<()> {
         unsafe {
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @superboum.

notgull:
Should we test Haiku in the CI?

Thomasdezeeuw:
> Should we test Haiku in the CI?

Yes, but I don't think Haiku has (or had) a rustup target that would could easily download. If that has changed adding it to https://github.com/rust-lang/socket2/blob/ed233839be976d07f463ac3eb4dc0d6e19e333ab/Makefile#L2 and https://github.com/rust-lang/socket2/blob/ed233839be976d07f463ac3eb4dc0d6e19e333ab/.github/workflows/main.yml#L64 should add it to the CI.

Issue #370, opened 2022-12-24
Support TCP_CONGESTION socket option


BobAnkh:
`TCP_CONGESTION` is a socket option to choose which tcp congestion control algorithms to use for the socket, which is supported on [Linux(since 2.6.13)](https://man7.org/linux/man-pages/man7/tcp.7.html) and [FreeBSD(since 9.0)](https://www.freebsd.org/cgi/man.cgi?query=tcp&sektion=4).

I wonder would you like to have this socket option in this crate? If so, I would like to contribute this since I have tried to implement it.

While implementing this feature, a string-like(here I choose `&str`) algorithm name should be passed to the setsockopt function(for example, "reno"). 
However, while simply convert the name to `&[u8]` and pass it to the wrapper function `setsockopt`, the inner calculation of length for this(`mem::size_of::<T>`) can not have the right value(i.e., true length is 4 while mem::size_of::<T> gives 16). 
Thus, I now implement it with raw `syscall!` and I wonder if there is a better way to do this?

Thomasdezeeuw:
> `TCP_CONGESTION` is a socket option to choose which tcp congestion control algorithms to use for the socket, which is supported on [Linux(since 2.6.13)](https://man7.org/linux/man-pages/man7/tcp.7.html) and [FreeBSD(since 9.0)](https://www.freebsd.org/cgi/man.cgi?query=tcp&sektion=4).
> 
> I wonder would you like to have this socket option in this crate? If so, I would like to contribute this since I have tried to implement it.

This addition would be welcome.

> While implementing this feature, a string-like(here I choose `&str`) algorithm name should be passed to the setsockopt function(for example, "reno"). However, while simply convert the name to `&[u8]` and pass it to the wrapper function `setsockopt`, the inner calculation of length for this(`mem::size_of::<T>`) can not have the right value(i.e., true length is 4 while mem::size_of:: gives 16). Thus, I now implement it with raw `syscall!` and I wonder if there is a better way to do this?

`mem::size_of` returns the memory size of the a type, in this case `&str` and `&[u8]` both are a pointer and a length, so 2 * 8 bytes = 16 bytes on 64 bit. What you're looking for is the length of the string/bytes which you can get by calling `.len()` on the string/byte slice.

However I don't think you want to use `&str` or `&[u8]` but [`std::ffi::CStr`](https://doc.rust-lang.org/std/ffi/struct.CStr.html), since C expects a C string (i.e. null terminated).

[...]

PR #371, opened 2022-12-24
Add support for TCP_CONGESTION socketopt

BobAnkh:
- `tcp_congestion` uses raw syscall since diff os have diff policy on modifying len
- `set_tcp_congestion` uses raw syscall since exact str len must be passed in

Closes: #370 

Co-authored-by: Campbell He <kp.campbell.he@duskmoon314.com>

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -2171,21 +2171,21 @@ impl crate::Socket {
     pub fn tcp_congestion(&self) -> io::Result<String> {
         let mut payload: MaybeUninit<[u8; TCP_CA_NAME_MAX]> = MaybeUninit::uninit();
         let mut len = size_of::<[u8; TCP_CA_NAME_MAX]>() as libc::socklen_t;
-        unsafe {
-            syscall!(getsockopt(
-                self.as_raw(),
-                IPPROTO_TCP,
-                libc::TCP_CONGESTION,
-                payload.as_mut_ptr().cast(),
-                &mut len,
-            ))
-            .map(|_| payload.assume_init())
-            .map(|buf| {
-                String::from_utf8_lossy(&buf)
-                    .trim_matches(char::from(0))
-                    .to_string()
-            })
-        }
+        syscall!(getsockopt(
+            self.as_raw(),
+            IPPROTO_TCP,
+            libc::TCP_CONGESTION,
+            payload.as_mut_ptr().cast(),
+            &mut len,
+        ))
+        .map(|_| {
+            let buf = unsafe { payload.assume_init() };
+            let raw_name = String::from_utf8_lossy(&buf);
+            match raw_name.split_once(char::from(0)) {
+                Some((left, _)) => left.to_string(),
+                None => raw_name.to_string(),
+            }
+        })
     }
 
     /// Set the value of the `TCP_CONGESTION` option for this socket.
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -2169,8 +2169,8 @@ impl crate::Socket {
         doc(cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux"))))
     )]
     pub fn tcp_congestion(&self) -> io::Result<Vec<u8>> {
-        let mut payload: MaybeUninit<[u8; TCP_CA_NAME_MAX]> = MaybeUninit::uninit();
-        let mut len = size_of::<[u8; TCP_CA_NAME_MAX]>() as libc::socklen_t;
+        let mut payload: [u8; TCP_CA_NAME_MAX] = [0; TCP_CA_NAME_MAX];
+        let mut len = payload.len() as libc::socklen_t;
         syscall!(getsockopt(
             self.as_raw(),
             IPPROTO_TCP,
@@ -2179,8 +2179,7 @@ impl crate::Socket {
             &mut len,
         ))
         .map(|_| {
-            let buf = unsafe { payload.assume_init() };
-            let buf = &buf[..len as usize];
+            let buf = &payload[..len as usize];
             // TODO: use `MaybeUninit::slice_assume_init_ref` once stable.
             unsafe { &*(buf as *const [_] as *const [u8]) }.into()
         })
@@ -2202,7 +2201,7 @@ impl crate::Socket {
             self.as_raw(),
             IPPROTO_TCP,
             libc::TCP_CONGESTION,
-            tcp_ca_name.as_ptr() as *const std::os::raw::c_void,
+            tcp_ca_name.as_ptr() as *const _,
             tcp_ca_name.len() as libc::socklen_t,
         ))
         .map(|_| ())
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -2168,7 +2168,7 @@ impl crate::Socket {
         docsrs,
         doc(cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux"))))
     )]
-    pub fn tcp_congestion(&self) -> io::Result<String> {
+    pub fn tcp_congestion(&self) -> io::Result<Vec<u8>> {
         let mut payload: MaybeUninit<[u8; TCP_CA_NAME_MAX]> = MaybeUninit::uninit();
         let mut len = size_of::<[u8; TCP_CA_NAME_MAX]>() as libc::socklen_t;
         syscall!(getsockopt(
@@ -2180,11 +2180,8 @@ impl crate::Socket {
         ))
         .map(|_| {
             let buf = unsafe { payload.assume_init() };
-            let raw_name = String::from_utf8_lossy(&buf);
-            match raw_name.split_once(char::from(0)) {
-                Some((left, _)) => left.to_string(),
-                None => raw_name.to_string(),
-            }
+            let name = buf.splitn(2, |num| *num == 0).next().unwrap().to_vec();
+            name
         })
     }
 
@@ -2199,14 +2196,13 @@ impl crate::Socket {
         docsrs,
         doc(cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux"))))
     )]
-    pub fn set_tcp_congestion(&self, tcp_ca_name: &str) -> io::Result<()> {
-        let name = std::ffi::OsString::from(tcp_ca_name);
+    pub fn set_tcp_congestion(&self, tcp_ca_name: &[u8]) -> io::Result<()> {
         syscall!(setsockopt(
             self.as_raw(),
             IPPROTO_TCP,
             libc::TCP_CONGESTION,
-            name.as_bytes().as_ptr() as *const std::os::raw::c_void,
-            name.len() as libc::socklen_t,
+            tcp_ca_name.as_ptr() as *const std::os::raw::c_void,
+            tcp_ca_name.len() as libc::socklen_t,
         ))
         .map(|_| ())
     }
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -2169,13 +2169,22 @@ impl crate::Socket {
         doc(cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux"))))
     )]
     pub fn tcp_congestion(&self) -> io::Result<String> {
+        let mut payload: MaybeUninit<[u8; TCP_CA_NAME_MAX]> = MaybeUninit::uninit();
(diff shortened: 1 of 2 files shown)


BobAnkh:
Currently have no idea why FreeBSD throws an error on the code `debug_assert_eq!(len as usize, size_of::<T>());` since `let mut len = size_of::<T>() as libc::socklen_t;`.

Woking on this now.

BobAnkh:
I have fixed the error thrown by FreeBSD. This is because that FreeBSD will modify the len field to the true string lenght while Linux not.

Now this pr is ready for review.

duskmoon314 (approved):

src/sys/unix.rs
```diff
@@ -2154,6 +2157,59 @@ impl crate::Socket {
...
+    ///
+    /// For more information about this option, see [`set_tcp_congestion`].
+    ///
+    /// [`set_tcp_congestion`]: Socket::set_tcp_congestion
+    #[cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux")))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux"))))
+    )]
+    pub fn tcp_congestion(&self) -> io::Result<String> {
```
Thomasdezeeuw, line 2171: Nothing in the documentation states that this always returns valid UTF-8, so we shouldn't return a `String` here, but `Vec<u8>` instead.

src/sys/unix.rs
```diff
@@ -2154,6 +2157,59 @@ impl crate::Socket {
...
+    ///
+    /// The value must be a valid TCP congestion control algorithm name of the
+    /// platform. For example, Linux may supports "reno", "cubic".
+    #[cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux")))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux"))))
+    )]
+    pub fn set_tcp_congestion(&self, tcp_ca_name: &str) -> io::Result<()> {
+        let name = std::ffi::OsString::from(tcp_ca_name);
```
Thomasdezeeuw, line 2203: Why are you making an allocation, only to use it as pointer? You can use `tcp_ca_name.as_ptr()`.

src/sys/unix.rs
```diff
@@ -2154,6 +2157,59 @@ impl crate::Socket {
...
+    /// Specifies the TCP congestion control algorithm to use for this socket.
+    ///
+    /// The value must be a valid TCP congestion control algorithm name of the
+    /// platform. For example, Linux may supports "reno", "cubic".
+    #[cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux")))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux"))))
+    )]
+    pub fn set_tcp_congestion(&self, tcp_ca_name: &str) -> io::Result<()> {
```
Thomasdezeeuw, line 2202: Let's change this to `&[u8]`.

Thomasdezeeuw (changes requested):

src/sys/unix.rs
BobAnkh: Done

src/sys/unix.rs
BobAnkh: Done

BobAnkh:
Code has been polished again, ready for another review!

src/sys/unix.rs
```diff
@@ -2154,6 +2157,55 @@ impl crate::Socket {
...
+        let mut len = size_of::<[u8; TCP_CA_NAME_MAX]>() as libc::socklen_t;
+        syscall!(getsockopt(
+            self.as_raw(),
+            IPPROTO_TCP,
+            libc::TCP_CONGESTION,
+            payload.as_mut_ptr().cast(),
+            &mut len,
+        ))
+        .map(|_| {
+            let buf = unsafe { payload.assume_init() };
```
Thomasdezeeuw, line 2182: This is UB. Only `&buf[0..len]` bytes are initialised by the kernel.

src/sys/unix.rs
```diff
@@ -2154,6 +2157,55 @@ impl crate::Socket {
...
+            let name = buf.splitn(2, |num| *num == 0).next().unwrap().to_vec();
+            name
```
Thomasdezeeuw, line 2184: I think the following will work.
```suggestion
            let buf = &buf[..len as usize];
            // TODO: use `MaybeUninit::slice_assume_init_ref` once stable.
            Ok(Some(unsafe { &*(buf as *const [_] as *const [u8]) }.into()))
```

tests/socket.rs
```diff
@@ -1341,3 +1341,25 @@ fn original_dst_ipv6() {
...
+    let origin_tcp_ca = socket
+        .tcp_congestion()
+        .expect("failed to get tcp congestion algorithm");
+    socket
+        .set_tcp_congestion(&origin_tcp_ca)
+        .expect("failed to set tcp congestion algorithm");
+    // Return a Err when set a non-exist tcp_ca
+    socket
+        .set_tcp_congestion(b"tcp_congestion_does_not_exist")
+        .unwrap_err();
```
Thomasdezeeuw, line 1359: Can you also try setting a different algorithm. I think "cubic" and "reno" are pretty common, you can try those two.

Thomasdezeeuw (changes requested):

tests/socket.rs
BobAnkh: I will try to set them. The original reason I do not explicitly set a algorithm is that the default ca for FreeBSD is newreno, not sure if others are available. But let's give it a try on CI. If not, I think it is better to split the test for different os if you agree.

src/sys/unix.rs
BobAnkh: Here do you mean directly return the [u8; len](in vec format)? This may have more than one `\0` appreared in the Vec<u8> returned.

E.g., In Linux, tcp_ca name is `bbr` and the return value is `[98, 98, 114, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]`.

If this is intended, than I will do as you suggest.

duskmoon314: If the desired behavior is the user need to parse `[u8]` to some string themselves, then maybe not splitting the buffer is better.

tests/socket.rs
BobAnkh: According to the documentation of FreeBSD, it only supports cubic not reno(since it uses newreno). Do we only set "cubic" in the test or split the test according to os?

BobAnkh: Ok, I will implement like this.

BobAnkh:
Some modification:

- New tcp ca test is added
- tcp_congestion return the untruncated u8 buf[..len]

BobAnkh:
Execute the command `sysctl net.inet.tcp.cc.available` on FreeBSD CI, which shows that it only supports `newreno`. This might be because other cc modules are not loaded.

image

So now I only test `newreno` on FreeBSD. Any suggestion?

BobAnkh:
I have a suggestion that can we use OsString instead as the crate `nix` does? Accepting a &[u8] and especially returning the Vec<u8> with trailing `\0`s seems not so user-friendly I'm afraid. Up to you to confirm which interface you prefer.

Also, `newreno` is the only ca we can test on CI's FreeBSD.

So any ideas on the two points? Or is there anything else I should implement?

Thomasdezeeuw:
> So now I only test `newreno` on FreeBSD. Any suggestion?

That's fine, let's just set `newreno` again.

> I have a suggestion that can we use OsString instead as the crate `nix` does? Accepting a &[u8] and especially returning the Vec with trailing `\0`s seems not so user-friendly I'm afraid. Up to you to confirm which interface you prefer.

No, we're already using `Vec<u8>`/`&[u8]` other places. Let's stick with a consistent API.

BobAnkh:
Get  your point. Then I've implemented all requested.

src/sys/unix.rs
```diff
@@ -204,6 +204,9 @@ const MAX_BUF_LEN: usize = ssize_t::MAX as usize;
 #[cfg(target_vendor = "apple")]
 const MAX_BUF_LEN: usize = c_int::MAX as usize - 1;
 
+#[cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux")))]
+const TCP_CA_NAME_MAX: usize = 16;
```
Thomasdezeeuw, line 208: Let's add this constant to libc.

src/sys/unix.rs
```diff
@@ -2154,6 +2157,56 @@ impl crate::Socket {
...
+    ///
+    /// [`set_tcp_congestion`]: Socket::set_tcp_congestion
+    #[cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux")))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux"))))
+    )]
+    pub fn tcp_congestion(&self) -> io::Result<Vec<u8>> {
+        let mut payload: MaybeUninit<[u8; TCP_CA_NAME_MAX]> = MaybeUninit::uninit();
+        let mut len = size_of::<[u8; TCP_CA_NAME_MAX]>() as libc::socklen_t;
```
Thomasdezeeuw, line 2173: This should really use the length of the payload. Although that's the same.
```suggestion
        let mut len = payload.len();
```

src/sys/unix.rs
Thomasdezeeuw: This is not resolved. You can't call `payload.assume_init`, it's not **fully** initialised. I think you can simply remove this line and we're good.

src/sys/unix.rs
```diff
@@ -2154,6 +2157,56 @@ impl crate::Socket {
...
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(feature = "all", any(target_os = "freebsd", target_os = "linux"))))
+    )]
+    pub fn set_tcp_congestion(&self, tcp_ca_name: &[u8]) -> io::Result<()> {
+        syscall!(setsockopt(
+            self.as_raw(),
+            IPPROTO_TCP,
+            libc::TCP_CONGESTION,
+            tcp_ca_name.as_ptr() as *const std::os::raw::c_void,
```
Thomasdezeeuw, line 2205: Nit: This way we don't have to worry about different types on different OS.
```suggestion
            tcp_ca_name.as_ptr() as *const _,
```

tests/socket.rs
```diff
@@ -1341,3 +1341,50 @@ fn original_dst_ipv6() {
...
+    let cur_tcp_ca = socket.tcp_congestion().unwrap();
+    assert_eq!(
+        cur_tcp_ca, origin_tcp_ca,
+        "expected {origin_tcp_ca:?} but get {cur_tcp_ca:?}"
```
Thomasdezeeuw, line 1363: Not needed, `assert_eq` prints it already.
```suggestion
```

Thomasdezeeuw (changes requested):
Still has UB.

src/sys/unix.rs
BobAnkh: Hey, `TCP_CA_NAME_MAX` isn't defined in user space include files, this is a kernel-side definition. So what should I do?

Thomasdezeeuw: Can you add a comment saying so, I prefer to use `libc`'s constant, but if that's not possible it's ok to define it ourselves.

src/sys/unix.rs
Thomasdezeeuw: Not what I asked for, but fine for now.

Allow me to explain the UB in the old code one more time. You create an array of size `TCP_CA_NAME_MAX`, accessing any of those by would be UB. Then you call the kernel to fill the bytes, but it doesn't fill all of them just `n` bytes (return by the system call). With the old code you called `payload.assume_init()`, which requires the entire array all (`TCP_CA_NAME_MAX` bytes) to be initialised, but that's not always the case as it's only up to `n` bytes.

So the only change needed to the old code was to remove the `payload.assume_init()` line. Because the next line creating a slice of only the initialised bytes and the line after that cast it to a slice of bytes (`MaybeUninit<u8>` -> `u8`).

Thomasdezeeuw (approved):

src/sys/unix.rs
BobAnkh: Ok I will add the comment.

Thomasdezeeuw:
Thanks @BobAnkh

Issue #372, opened 2022-12-30
Implement `std::cmp::Eq` and `std::hash::Hash` for `socket2::SockAddr`, so it can be used as a key in `std::collections::HashMap`


lshlyapnikov:

Thomasdezeeuw:
I'm not sure about `Hash`, but I think we can implement `Eq`. But do note that `SockAddr` is quite big, if you're using just IP address you're better of using `std::net::SocketAddr`.

lshlyapnikov:
@Thomasdezeeuw you mean it is not possible or are there any other reasons?

My use case is tracking UDP packet sequence numbers per sender (`socket2::SockAddr`).
I use `recv_from` which gives me the `SockAddr` of the sender.
```
pub fn recv_from(&self, buf: &mut [MaybeUninit<u8>]) -> io::Result<(usize, SockAddr)>
```
and a `HashMap` to keep expected sequence numbers per sender.

I currently have to implement a newtype that wraps around `socket2::SockAddr` and implement `PartialEq` and `Hash` for it.

I guess can I can open a pull request. Let me know.

lshlyapnikov:
you know... it is actually for IP address only and `std::net::SocketAddr` has `Hash` and `Eq` implemented. I could use ``std::net::SocketAddr` as a key in `HashMap`.  However converting from `socket2::SockAddr` to `std::net::SocketAddr` is not free and I don't want to do this for every UDP packet I receive.
```
pub fn as_socket(&self) -> Option<SocketAddr> {
```

Thomasdezeeuw:
> It is actually for IP address only and `std::net::SocketAddr` has `Hash` and `Eq` implemented. I could use `std::net::SocketAddr` as a key in `HashMap`. However converting from `socket2::SockAddr` to `std::net::SocketAddr` is not free and I don't want to do this for every UDP packet I receive.
> 
> ```
> pub fn as_socket(&self) -> Option<SocketAddr>
> ```

Using `std::net::SocketAddr` reduce the memory usage, so I think it would worth benchmarking both. The conversion to and from `socket2::SockAddr` isn't too costly I think.

But a pr to implement it would be welcome.

lshlyapnikov:
See #373. I will ping you when it is ready for review.

PR #373, opened 2023-01-02
372: Implement `Eq` and `Hash` for `socket2::SockAddr`

lshlyapnikov:
Closes: #372.

Cannot derive `Eq` and `Hash` for `socket2::SockAddr` because `sockaddr_storage` does not have both traits derived/implemented. `sockaddr_storage` is defined within `s_no_extra_traits`, my assumption it is expected to stay this way and I should not derive/implement `Eq` and `Hash` for `sockaddr_storage`.

## TODO:
- [x] add test cases
- [ ] add benchmarks for `socket2::SockAddr::as_socket` + `std::net::SocketAddr::hash` vs `socket2::SockAddr::hash` (this implementation)
- [ ] target this at master, v0.4 is only accepting backports.

### DIFF
--- a/src/sockaddr.rs
+++ b/src/sockaddr.rs
@@ -1,3 +1,4 @@
+use std::hash::Hash;
 use std::mem::{self, size_of, MaybeUninit};
 use std::net::{SocketAddr, SocketAddrV4, SocketAddrV6};
 use std::{fmt, io};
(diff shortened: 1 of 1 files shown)


src/sockaddr.rs
```diff
@@ -310,40 +310,192 @@ impl fmt::Debug for SockAddr {
...
-    assert!(addr.as_socket_ipv6().is_none());
-
-    let addr = SockAddr::from(SocketAddr::from(std));
-    assert_eq!(addr.family(), AF_INET as sa_family_t);
-    assert_eq!(addr.len(), size_of::<sockaddr_in>() as socklen_t);
-    assert_eq!(addr.as_socket(), Some(SocketAddr::V4(std)));
-    assert_eq!(addr.as_socket_ipv4(), Some(std));
-    assert!(addr.as_socket_ipv6().is_none());
+impl cmp::PartialEq for SockAddr {
+    #[inline]
```
Thomasdezeeuw, line 314: No need for these inline attributes.

src/sockaddr.rs
```diff
@@ -310,40 +310,192 @@ impl fmt::Debug for SockAddr {
...
+    fn eq(&self, other: &Self) -> bool {
+        let this_sockaddr = self.as_ptr();
+        let other_sockaddr = other.as_ptr();
+        unsafe {
+            (*this_sockaddr).sa_family == (*other_sockaddr).sa_family
+                && (*this_sockaddr).sa_data == (*other_sockaddr).sa_data
```
Thomasdezeeuw, line 320: This is UB for short address. Please look at the `as_socket` method and only access the initialised data.

src/sockaddr.rs
```diff
@@ -310,40 +310,192 @@ impl fmt::Debug for SockAddr {
...
+}
+
+impl cmp::Eq for SockAddr {}
+
+impl hash::Hash for SockAddr {
+    #[inline]
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        let this_sockaddr = self.as_ptr();
+        unsafe {
+            ((*this_sockaddr).sa_family, (*this_sockaddr).sa_data).hash(state);
```
Thomasdezeeuw, line 332: Same as for the `PartialEq` implementation.

src/sockaddr.rs
```diff
@@ -310,40 +310,192 @@ impl fmt::Debug for SockAddr {
...
-    assert_eq!(addr.as_socket_ipv4(), Some(std));
-    assert!(addr.as_socket_ipv6().is_none());
-
-    let addr = SockAddr::from(SocketAddr::from(std));
-    assert_eq!(addr.family(), AF_INET as sa_family_t);
-    assert_eq!(addr.len(), size_of::<sockaddr_in>() as socklen_t);
-    assert_eq!(addr.as_socket(), Some(SocketAddr::V4(std)));
-    assert_eq!(addr.as_socket_ipv4(), Some(std));
-    assert!(addr.as_socket_ipv6().is_none());
+impl cmp::PartialEq for SockAddr {
```
Thomasdezeeuw, line 313: I think `PartialEq` is in the prelude of std lib.

src/sockaddr.rs
```diff
@@ -310,40 +310,192 @@ impl fmt::Debug for SockAddr {
...
+        unsafe {
+            (*this_sockaddr).sa_family == (*other_sockaddr).sa_family
+                && (*this_sockaddr).sa_data == (*other_sockaddr).sa_data
+        }
+    }
+}
+
+impl cmp::Eq for SockAddr {}
+
+impl hash::Hash for SockAddr {
```
Thomasdezeeuw, line 327: You can just import `Hash`, no need for `hash::Hash`.

Thomasdezeeuw (changes requested):
Please target this at master, v0.4 is only accepting backports.

src/sockaddr.rs
lshlyapnikov: done

src/sockaddr.rs
lshlyapnikov: done

src/sockaddr.rs
lshlyapnikov: done

src/sockaddr.rs
lshlyapnikov: @Thomasdezeeuw 
What do I do in case `ss_family` is neither `AF_INET` nor `AF_INET6`... do I panic?
```
/// Returns this address as a `SocketAddr` if it is in the `AF_INET` (IPv4)
    /// or `AF_INET6` (IPv6) family, otherwise returns `None`.
    pub fn as_socket(&self) -> Option<SocketAddr> {
        if self.storage.ss_family == AF_INET as sa_family_t {
...
        } else if self.storage.ss_family == AF_INET6 as sa_family_t {
...
        } else {
            None
        }
    }
```

```
pub struct SockAddr {
    storage: sockaddr_storage,
    len: socklen_t,
}
```
we know the `len`gth of the `storage: sockaddr_storage`, can I simply compare raw bytes?

lshlyapnikov:
Closing this PR, because it was base off of `v0.4.x` branch by mistake. See #374, which is based off of `master`.

PR #374, opened 2023-01-02
372: Implement `Eq` and `Hash` for `socket2::SockAddr`

lshlyapnikov:
Closes: #372.

This PR started as: #373, which was based off of `v0.4.x` by mistake, changing the base to `master`.

Cannot derive `Eq` and `Hash` for `socket2::SockAddr` because `sockaddr_storage` does not have both traits derived/implemented. `sockaddr_storage` is defined within `s_no_extra_traits`, my assumption it is expected to stay this way and I should not derive/implement `Eq` and `Hash` for `sockaddr_storage`.

## TODO:
- [x] add test cases. Quickcheck tests removed from this PR, can be found here: https://github.com/lshlyapnikov/socket2/blob/sockaddr-quickcheck/src/sockaddr.rs#L501-L554
- [x] add benchmarks for `socket2::SockAddr::as_socket` + `std::net::SocketAddr::hash` vs `socket2::SockAddr::hash` (this implementation). Benchmarks are not part of this PR, can be found here: https://github.com/lshlyapnikov/socket2/blob/sockaddr-quickcheck/benches/sockaddr_bench.rs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,3 @@ features = [
 [features]
 # Enable all API, even ones not available on all OSs.
 all = []
-
-[dev-dependencies]
-quickcheck = "1.0.3"
-quickcheck_macros = "1.0.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,3 +48,7 @@ features = [
 [features]
 # Enable all API, even ones not available on all OSs.
 all = []
+
+[dev-dependencies]
+quickcheck = "1.0.3"
+quickcheck_macros = "1.0.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,12 +59,6 @@
 // Disallow warnings in examples.
 #![doc(test(attr(deny(warnings))))]
 
-#[cfg(test)]
-extern crate quickcheck;
-#[cfg(test)]
-#[macro_use(quickcheck)]
-extern crate quickcheck_macros;
-
 use std::fmt;
 use std::mem::MaybeUninit;
 use std::net::SocketAddr;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,6 +59,12 @@
 // Disallow warnings in examples.
 #![doc(test(attr(deny(warnings))))]
 
+#[cfg(test)]
+extern crate quickcheck;
+#[cfg(test)]
+#[macro_use(quickcheck)]
+extern crate quickcheck_macros;
+
 use std::fmt;
 use std::mem::MaybeUninit;
 use std::net::SocketAddr;
--- a/src/sockaddr.rs
+++ b/src/sockaddr.rs
@@ -341,13 +341,6 @@ unsafe fn any_as_u8_slice<T: Sized>(p: &T, size: usize) -> &[u8] {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use quickcheck::{self, Arbitrary};
-
-    impl Arbitrary for SockAddr {
-        fn arbitrary(g: &mut quickcheck::Gen) -> Self {
-            SockAddr::from(std::net::SocketAddr::arbitrary(g))
-        }
-    }
 
     #[test]
     fn ipv4() {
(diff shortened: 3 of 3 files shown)


src/sockaddr.rs
```diff
@@ -312,44 +313,196 @@ impl fmt::Debug for SockAddr {
...
-    assert_eq!(addr.len(), size_of::<sockaddr_in>() as socklen_t);
-    assert_eq!(addr.as_socket(), Some(SocketAddr::V4(std)));
-    assert_eq!(addr.as_socket_ipv4(), Some(std));
-    assert!(addr.as_socket_ipv6().is_none());
+impl PartialEq for SockAddr {
+    fn eq(&self, other: &Self) -> bool {
+        unsafe {
+            let these_bytes: &[u8] = any_as_u8_slice(&self.storage, self.len as usize);
+            let those_bytes: &[u8] = any_as_u8_slice(&other.storage, other.len as usize);
+            these_bytes == those_bytes
```
lshlyapnikov, line 321: @Thomasdezeeuw what do you think about this implementation? I am casting `storage` values to a byte slices and comparing them.

The approach that you suggested [here](https://github.com/rust-lang/socket2/pull/373#discussion_r1059961032) will require some special treatment for the case when `SockAddr` is neither `AF_INET` or `AF_INET6` and I don't have a good solution for this case.

lshlyapnikov: Also what do you think about adding quickcheck test dependency to the project? It is for the property based testing.

See https://github.com/BurntSushi/quickcheck

I could use it for more comprehensive testing of `Eq` and `Hash` implementations.

Cargo.toml
```diff
@@ -48,3 +48,7 @@ features = [
 [features]
 # Enable all API, even ones not available on all OSs.
 all = []
+
+[dev-dependencies]
```
Thomasdezeeuw, line 52: Please don't add (dev) dependencies.

src/lib.rs
```diff
@@ -59,6 +59,12 @@
 // Disallow warnings in examples.
 #![doc(test(attr(deny(warnings))))]
 
+#[cfg(test)]
+extern crate quickcheck;
+#[cfg(test)]
+#[macro_use(quickcheck)]
+extern crate quickcheck_macros;
```
Thomasdezeeuw, line 66: FYI: since rust 2018 this is no longer required.

src/sockaddr.rs
Thomasdezeeuw: Sorry, I don't have a lot of time to review this atm, so I just took a quick look.

> Uninitialized fields are set to `0` by the `From` implementations:
> 
> ```
> let mut storage = unsafe { mem::zeroed::<sockaddr_storage>() };
> ```

That's not the only way to create a `SockAddr`, so we can't depending on this being true. We simply cannot access storage beyond `len` bytes.

> `SockAddr::new` rustdoc advises to do the same before populating `sockaddr_storage` and passing it to the constructor.
> 
> The approach that you suggested [here](https://github.com/rust-lang/socket2/pull/373#discussion_r1059961032) will require some special treatment for the case when `SockAddr` is neither `AF_INET` nor `AF_INET6` and I don't have a good solution for this case. I can panic or return `false` which would not work for other protocol families/domains.

Please don't panic in you code, it's very rarely the right thing to do.

lshlyapnikov: yep... no panic here.

> We simply cannot access storage beyond `len` bytes.
and that is exactly what `any_as_u8_slice` does, it creates a byte slice limited by `len`, see the 2nd argument:
```
let these_bytes: &[u8] = any_as_u8_slice(&self.storage, self.len as usize);
let those_bytes: &[u8] = any_as_u8_slice(&other.storage, other.len as usize);
```

Cargo.toml
lshlyapnikov: you mean don't add `quickcheck` at all or add it ad a regular prod dependency? It is for testing only.

src/lib.rs
lshlyapnikov: `#[quickcheck]` attribute won't work without it: https://github.com/BurntSushi/quickcheck/blob/master/README.md#the-quickcheck-attribute

However if you don't want any new dependencies, this will go away.

Cargo.toml
Thomasdezeeuw: I mean either, socket2 is used by a lot of crates, introducing a new dependency takes to properly fed it, which I don't have a the moment. Furthermore the added befit is slim.

lshlyapnikov: quickcheck tests and dev dependency removed.

Thomasdezeeuw (approved):

lshlyapnikov:
Benchmarks comparing `socket2::SockAddr::hash` with calling `hash` on `socket2::SockAddr::as_socket`

https://github.com/lshlyapnikov/socket2/blob/sockaddr-quickcheck/benches/sockaddr_bench.rs

https://github.com/lshlyapnikov/socket2/blob/sockaddr-quickcheck/benches/sockaddr_bench.rs#L15-L18

https://github.com/lshlyapnikov/socket2/blob/sockaddr-quickcheck/benches/sockaddr_bench.rs#L15-L18

```
test _benchmark_as_socket_hash ... bench:          33 ns/iter (+/- 22)
test _benchmark_hash           ... bench:          18 ns/iter (+/- 6)
```

I will keep these benchmarks in my fork. Let me know when/if you want to add them to the `master`.

Thomasdezeeuw:
> Running benches/sockaddr_bench.rs (target/release/deps/sockaddr_bench-8434022fa4125c76)
> 
> running 2 tests
> test benchmark_as_socket_hash ... bench:          42 ns/iter (+/- 21)
> test benchmark_hash           ... bench:          19 ns/iter (+/- 8)
> ```

Looks pretty decent.

> I will keep these benchmarks in my fork. Let me know when/if you want to add them to the `master`.

I think that's ok for now. If we ever get a complaint about the performance we can add them.

Thanks for the pr.

PR #389, opened 2023-02-24
Add `Socket::peek_sender()`

abonander:
Motivation: https://github.com/tokio-rs/tokio/issues/5491

### DIFF
--- a/src/sys/windows.rs
+++ b/src/sys/windows.rs
@@ -469,6 +469,38 @@ pub(crate) fn recv_from(
     }
 }
 
+pub(crate) fn peek_sender(socket: Socket) -> io::Result<SockAddr> {
+    // Safety: `recvfrom` initialises the `SockAddr` for us.
+    let ((), sender) = unsafe {
+        SockAddr::try_init(|storage, addrlen| {
+            let res = syscall!(
+                recvfrom(
+                    socket,
+                    // Windows *appears* not to care if you pass a null pointer.
+                    ptr::null_mut(),
+                    0,
+                    MSG_PEEK,
+                    storage.cast(),
+                    addrlen,
+                ),
+                PartialEq::eq,
+                SOCKET_ERROR
+            );
+            match res {
+                Ok(_n) => Ok(()),
+                Err(e) => match e.raw_os_error() {
+                    Some(code) if code == (WSAESHUTDOWN as i32) || code == (WSAEMSGSIZE as i32) => {
+                        Ok(())
+                    }
+                    _ => Err(e),
+                },
+            }
+        })
+    }?;
+
+    Ok(sender)
+}
+
 pub(crate) fn recv_from_vectored(
     socket: Socket,
     bufs: &mut [crate::MaybeUninitSlice<'_>],
--- a/tests/socket.rs
+++ b/tests/socket.rs
@@ -532,6 +532,21 @@ fn out_of_band() {
 }
 
 #[test]
+#[cfg(not(target_os = "redox"))] // cfg of `udp_pair_unconnected()`
+fn udp_peek_sender() {
+    let (socket_a, socket_b) = udp_pair_unconnected();
+
+    let socket_a_addr = socket_a.local_addr().unwrap();
+    let socket_b_addr = socket_b.local_addr().unwrap();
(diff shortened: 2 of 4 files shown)


tests/socket.rs
```diff
@@ -531,6 +531,27 @@ fn out_of_band() {
...
+    socket_b.bind(&localhost_port_0).unwrap();
+
+    let socket_a_addr = socket_a.local_addr().unwrap();
+    let socket_b_addr = socket_b.local_addr().unwrap();
+
+    socket_b.send_to(b"Hello, world!", &socket_a_addr).unwrap();
+
+    let sender_addr = socket_a.peek_sender().unwrap();
+
+    assert_eq!(sender_addr.as_socket_ipv4(), socket_b_addr.as_socket_ipv4());
```
abonander, line 552: This is failing on macOS because `sender_addr` is somehow _not_ IPv4. 

Unless the answer here is obvious to everyone else but me, I'm gonna get set up on an iMac Pro at work tomorrow to hit this with the debugger and figure out what's actually being returned here. I'd just edit the test to print it out but it seems inconsiderate to use CI minutes for debugging.

src/sys/windows.rs
```diff
@@ -469,6 +469,38 @@ pub(crate) fn recv_from(
...
 }
 
+pub(crate) fn peek_sender(socket: Socket) -> io::Result<SockAddr> {
+    // Safety: `recvfrom` initialises the `SockAddr` for us.
+    let ((), sender) = unsafe {
+        SockAddr::try_init(|storage, addrlen| {
+            let res = syscall!(
+                recvfrom(
+                    socket,
+                    //
```
Thomasdezeeuw, line 479:
```suggestion
```

tests/socket.rs
Thomasdezeeuw: It seems it doesn't work on macOS. We get an empty `SockAddr`:

```
SockAddr {
    ss_len: 0,
    ss_family: 0,
    len: 0,
}
```

src/sys/windows.rs
abonander: Heh, I wanted to say something there because it looks wrong passing a null pointer, but I wasn't sure what. Windows apparently doesn't care.

tests/socket.rs
abonander: It may not like the zero-sized read buffer. I figured I'd pass something small but non-zero sized like 8 bytes.

tests/socket.rs
```diff
@@ -531,6 +531,27 @@ fn out_of_band() {
...
+#[test]
+fn peek_sender_udp() {
+    let socket_a = Socket::new(Domain::IPV4, Type::DGRAM, Some(Protocol::UDP)).unwrap();
+    let socket_b = Socket::new(Domain::IPV4, Type::DGRAM, Some(Protocol::UDP)).unwrap();
+
+    let localhost_port_0 = SockAddr::from(SocketAddrV4::new(Ipv4Addr::LOCALHOST, 0));
+
+    // Allocate local address/port pairs for both sockets.
+    socket_a.bind(&localhost_port_0).unwrap();
+    socket_b.bind(&localhost_port_0).unwrap();
```
abonander, line 543: I noticed that there's a function to do this setup in the file already.

tests/socket.rs
abonander: Fix it by passing a non-zero sized buffer, at least on macOS on ARM.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @abonander

abonander:
@Thomasdezeeuw do you know roughly how long until 0.5 is released? I could also put in a PR to backport this if you like.

Thomasdezeeuw:
> @Thomasdezeeuw do you know roughly how long until 0.5 is released? I could also put in a PR to backport this if you like.

Don't really have an ETA, but I think the TODO list is shrinking. I'll see if I can release it this weekend, otherwise we can backport.

abonander:
The MSRV bump in 0.5 is unfortunately unacceptable for Tokio so it looks like I'll be backporting this anyway.

PR #401, opened 2023-02-26
Set rust-version in Cargo.toml

atouchet:
Now that the MSRV is higher than 1.56 it should be safe to set rust-version in Cargo.toml.


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @atouchet

Issue #332, opened 2022-08-19
Expose address from SockAddr for unix domains


sullivan-sean:
It would be nice to be able to access the path of a unix socket from `SockAddr` similar to `vsock_address`

`socket2` is being considered as a backend for UDS on windows in `mio`/`tokio` (https://github.com/tokio-rs/mio/pull/1610#issuecomment-1218388105) and, if used, would need to expose the address of unix sockets from `SocketAddr` to mirror the implementation of [SocketAddr](https://docs.rs/mio/latest/mio/net/struct.SocketAddr.html#method.as_pathname) in `mio` for unix

Thomasdezeeuw:
> It would be nice to be able to access the path of a unix socket from `SockAddr` similar to `vsock_address`

Do you mean https://docs.rs/socket2/latest/socket2/struct.SockAddr.html#method.unix? I think we simply forgot to expose it in https://github.com/rust-lang/socket2/pull/249.

> `socket2` is being considered as a backend for UDS on windows in `mio`/`tokio` ([tokio-rs/mio#1610 (comment)](https://github.com/tokio-rs/mio/pull/1610#issuecomment-1218388105)) and, if used, would need to expose the address of unix sockets from `SocketAddr` to mirror the implementation of [SocketAddr](https://docs.rs/mio/latest/mio/net/struct.SocketAddr.html#method.as_pathname) in `mio` for unix

It's not ;)

sullivan-sean:
I meant more going the other way from `SockAddr` -> `Path`. Right now the `SockAddr` struct has private fields and there are no methods for Unix sockets exposing the `Path` which means once you have a `SockAddr` object (e.g. as returned from `accept`) you cannot get the corresponding `Path` back out, or tell whether a `SockAddr` is an unnamed or abstract Unix socket (i.e. the `SockAddr` here can't provide this functionality https://github.com/tokio-rs/mio/blob/master/src/sys/unix/uds/socketaddr.rs#L76-L96)

Thomasdezeeuw:
We could add `SocketAddr::as_pathname() -> Option<Path>`.

sullivan-sean:
What is the desired behavior for abstract unix sockets where there is a name associated with the `SockAddr` struct but this name has no connection with filesystem  pathnames? (https://manpages.ubuntu.com/manpages/trusty/man7/unix.7.html)

Unnamed unix sockets would presumably return `None` from `as_pathname`, and normal sockets would return a `Path` as expected.

[...]

PR #403, opened 2023-02-27
Match the standard library's interface for unix sockets on unix.

t4lz:
Following #332, currently unix only. Should this interface also be made available on Windows?

This PR adds to `SockAddr` methods to get the path of the unix socket address on unix.
It also adds the new methods to the tests.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -666,8 +666,8 @@ impl SockAddr {
         }
     }
 
-    /// Returns true if this address is an unnamed address from the `AF_UNIX` (AKA `AF_LOCAL`)
-    /// family (for local interprocess communication), false otherwise.
+    /// Returns true if this address is an unnamed address from the `AF_UNIX` family (for local
+    /// interprocess communication), false otherwise.
     pub fn is_unnamed(&self) -> bool {
         self.as_sockaddr_un()
             .map(|storage| {
@@ -684,8 +684,8 @@ impl SockAddr {
     /// otherwise returns `None`.
     pub(crate) fn as_sockaddr_un(&self) -> Option<&libc::sockaddr_un> {
         self.is_unix().then(|| {
-            // SAFETY: if local, i.e. the `ss_family` field is `AF_UNIX` then storage must be a
-            // `sockaddr_un`.
+            // SAFETY: if unix socket, i.e. the `ss_family` field is `AF_UNIX` then storage must be
+            // a `sockaddr_un`.
             unsafe { &*self.as_ptr().cast::<libc::sockaddr_un>() }
         })
     }
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -569,6 +569,14 @@ impl<'a> MaybeUninitSlice<'a> {
     }
 }
 
+/// Returns the offset of the `sun_path` member of the passed unix socket address.
+pub(crate) fn offset_of_path(storage: &libc::sockaddr_un) -> usize {
+    let base = storage as *const _ as usize;
+    let path = ptr::addr_of!(storage.sun_path) as usize;
+    path - base
+}
+
+
 #[allow(unsafe_op_in_unsafe_fn)]
 pub(crate) fn unix_sockaddr(path: &Path) -> io::Result<SockAddr> {
     // SAFETY: a `sockaddr_storage` of all zeros is valid.
@@ -603,9 +611,7 @@ pub(crate) fn unix_sockaddr(path: &Path) -> io::Result<SockAddr> {
             );
         }
 
-        let base = storage as *const _ as usize;
-        let path = ptr::addr_of!(storage.sun_path) as usize;
-        let sun_path_offset = path - base;
+        let sun_path_offset = offset_of_path(storage);
         sun_path_offset
             + bytes.len()
             + match bytes.first() {
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -59,6 +59,7 @@ use std::path::Path;
 use std::ptr;
 use std::time::{Duration, Instant};
 use std::{io, slice};
+use std::ffi::OsStr;
 
 #[cfg(not(any(
     target_os = "ios",
@@ -95,7 +96,7 @@ pub(crate) use libc::IPPROTO_SCTP;
 pub(crate) use libc::{IPPROTO_ICMP, IPPROTO_ICMPV6, IPPROTO_TCP, IPPROTO_UDP};
 // Used in `SockAddr`.
 pub(crate) use libc::{
-    sa_family_t, sockaddr, sockaddr_in, sockaddr_in6, sockaddr_un, sockaddr_storage, socklen_t,
+    sa_family_t, sockaddr, sockaddr_in, sockaddr_in6, sockaddr_storage, socklen_t,
 };
 // Used in `RecvFlags`.
 #[cfg(not(target_os = "redox"))]
@@ -570,7 +571,7 @@ impl<'a> MaybeUninitSlice<'a> {
 }
 
 /// Returns the offset of the `sun_path` member of the passed unix socket address.
-pub(crate) fn offset_of_path(storage: &sockaddr_un) -> usize {
+pub(crate) fn offset_of_path(storage: &libc::sockaddr_un) -> usize {
     let base = storage as *const _ as usize;
     let path = ptr::addr_of!(storage.sun_path) as usize;
     path - base
@@ -582,7 +583,7 @@ pub(crate) fn unix_sockaddr(path: &Path) -> io::Result<SockAddr> {
     // SAFETY: a `sockaddr_storage` of all zeros is valid.
     let mut storage = unsafe { mem::zeroed::<sockaddr_storage>() };
     let len = {
-        let storage = unsafe { &mut *ptr::addr_of_mut!(storage).cast::<sockaddr_un>() };
+        let storage = unsafe { &mut *ptr::addr_of_mut!(storage).cast::<libc::sockaddr_un>() };
 
         let bytes = path.as_os_str().as_bytes();
         let too_long = match bytes.first() {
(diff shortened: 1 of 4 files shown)


src/sys/unix.rs
```diff
@@ -659,6 +665,75 @@ impl SockAddr {
...
+            .unwrap_or_default()
+    }
+
+    /// Returns the underlying `sockaddr_un` object if this addres is from the `AF_UNIX` family,
+    /// otherwise returns `None`.
+    pub(crate) fn as_sockaddr_un(&self) -> Option<&libc::sockaddr_un> {
+        self.is_unix().then(|| {
+            // SAFETY: if local, i.e. the `ss_family` field is `AF_UNIX` then storage must be a
+            // `sockaddr_un`.
+            unsafe { &*ptr::addr_of!(self.storage).cast::<libc::sockaddr_un>() }
```
aviramha, line 689: why not use `self.as_storage_ptr`? I assume that's why they added this?

t4lz: `as_storage_ptr` is not defined on `redox`.

This expression is what's used in the existing code to get a reference to the appropriate socket address type:

https://github.com/rust-lang/socket2/blob/18a883653752116a10ec78f7d36f8d956e4c2e7c/src/sockaddr.rs#L199

https://github.com/rust-lang/socket2/blob/18a883653752116a10ec78f7d36f8d956e4c2e7c/src/sockaddr.rs#L206

https://github.com/rust-lang/socket2/blob/18a883653752116a10ec78f7d36f8d956e4c2e7c/src/sys/unix.rs#L577

aviramha: hmm but that requires the visibility change that might be there to avoid using stuff on wrong platform.

aviramha: p.s `as_ptr` might work, no?

t4lz: Changed to call `as_ptr` instead of calling `addr_of!` directly.

src/sys/unix.rs
```diff
@@ -659,6 +665,75 @@ impl SockAddr {
...
+        })
+    }
+
+    /// Returns this address as a slice of bytes representing an abstract address if it is an
+    /// `AF_UNIX` abstract address, otherwise returns `None`.
+    ///
+    /// Abstract addresses are a Linux extension, so this method returns None on all non-Linux
+    /// platforms.
+    pub fn as_abstract_namespace(&self) -> Option<&[u8]> {
+        #[cfg(any(target_os = "linux", target_os = "android"))]
```
Thomasdezeeuw, line 727: Also needs fuchsia here (and below).

src/sys/unix.rs
```diff
@@ -659,6 +665,75 @@ impl SockAddr {
...
+        {
+            self.as_sockaddr_un().and_then(|storage| {
+                (self.len() > offset_of_path(storage) as u32 && storage.sun_path[0] == 0)
+                    .then(|| self.path_bytes(storage, false))
```
Thomasdezeeuw, line 731: I think this should skip the first byte.

tests/socket.rs
```diff
@@ -144,17 +146,48 @@ fn socket_address_unix() {
...
 fn socket_address_unix_abstract_namespace() {
     let path = "\0h".repeat(108 / 2);
-    let addr = SockAddr::unix(path).unwrap();
+    let addr = SockAddr::unix(&path).unwrap();
     assert_eq!(
         addr.len() as usize,
         std::mem::size_of::<libc::sockaddr_un>()
     );
+    assert!(!addr.is_unnamed());
+    assert_eq!(addr.as_abstract_namespace(), Some(path.as_bytes()));
```
Thomasdezeeuw, line 188: I don't think we shouldn't return the null byte.
```suggestion
    assert_eq!(addr.as_abstract_namespace(), Some(b"h"));
```

src/sys/unix.rs
```diff
@@ -659,6 +665,75 @@ impl SockAddr {
...
+
+    /// Returns true if this address is an unnamed address from the `AF_UNIX` (AKA `AF_LOCAL`)
+    /// family (for local interprocess communication), false otherwise.
+    pub fn is_unnamed(&self) -> bool {
+        self.as_sockaddr_un()
+            .map(|storage| {
+                self.len() == offset_of_path(storage) as u32
+                    // On some non-linux platforms a zeroed path is returned for unnamed.
+                    // Abstract addresses only exist on Linux.
+                    || (cfg!(not(any(target_os = "linux", target_os = "android")))
```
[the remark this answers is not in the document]
t4lz: Added fuchsia in https://github.com/rust-lang/socket2/pull/403/commits/8375cea06dc3ca108bda593ff6b8d87a82666d8c, although I couldn't really find any source that says whether Fuchsia has abstract-name unix sockets, and the existing code here did not include Fuchsia in platforms that have them:

https://github.com/rust-lang/socket2/blob/18a883653752116a10ec78f7d36f8d956e4c2e7c/tests/socket.rs#L150-L151

And neither does the standard library:
https://doc.rust-lang.org/src/std/os/unix/net/addr.rs.html#236

src/sys/unix.rs
```diff
@@ -659,6 +665,75 @@ impl SockAddr {
...
+    }
+
+    /// Get the length of the path bytes of the address, not including any terminating null.
+    fn path_len(&self, storage: &libc::sockaddr_un, null_terminated: bool) -> usize {
+        self.len() as usize - offset_of_path(storage) - if null_terminated { 1 } else { 0 }
+    }
+
+    /// Get a u8 slice for the bytes of the pathname or abstract name.
+    fn path_bytes(&self, storage: &libc::sockaddr_un, null_terminated: bool) -> &[u8] {
+        let path_len = self.path_len(storage, null_terminated);
```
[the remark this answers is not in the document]
t4lz: Done.

src/sys/unix.rs
t4lz: Right, thank you. Changed inside `path_bytes`.

src/sys/unix.rs
```diff
@@ -659,6 +665,87 @@ impl SockAddr {
...
+        // the last field in the storage and so its length is equal to
+        //          TOTAL_LENGTH - OFFSET_OF_PATH -1
+        // Where the 1 is either a terminating null if we have a pathname address, or the initial
+        // null byte, if it's an abstract name address. In the latter case, the path bytes start
+        // after the initial null byte, hence the `offset`.
+        // There is no safe way to convert a `&[i8]` ot `&[u8]`
+        unsafe {
+            slice::from_raw_parts(
+                (storage.sun_path.as_ptr() as *const u8).offset(abstract_name as isize),
+                self.path_len(storage),
```
Thomasdezeeuw, line 717: Not sure if this will work, but if we skip the first byte we need to reflect it in the length is as well. It might however be easier to ignore abstract names here and handle it in `as_abstract_namespace` by simply doing `path_bytes()[1..]` as we check for the first null byte there already.

```suggestion
                self.path_len(storage) - abstract_name as usize,
```

src/sys/unix.rs
```diff
@@ -659,6 +665,87 @@ impl SockAddr {
...
+    /// Should not be called on unnamed addresses.
+    fn path_bytes(&self, storage: &libc::sockaddr_un, abstract_name: bool) -> &[u8] {
+        debug_assert!(!self.is_unnamed());
+        // SAFETY: the pointed objects of type `i8` have the same memory layout as `u8`. The path is
+        // the last field in the storage and so its length is equal to
+        //          TOTAL_LENGTH - OFFSET_OF_PATH -1
+        // Where the 1 is either a terminating null if we have a pathname address, or the initial
+        // null byte, if it's an abstract name address. In the latter case, the path bytes start
+        // after the initial null byte, hence the `offset`.
+        // There is no safe way to convert a `&[i8]` ot `&[u8]`
```
Thomasdezeeuw, line 713:
```suggestion
        // There is no safe way to convert a `&[i8]` to `&[u8]`
```

src/sys/unix.rs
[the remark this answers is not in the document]
Thomasdezeeuw: @tamird can you confirm (or deny) that Fuchsia has abstract namespaces?

src/sys/unix.rs
t4lz: Already subtracting 1 from the length inside `path_len` as we need to subtract 1 anyways. Whether it's a pathname or an abstract name, there is a null byte (either at the end or at the start).

src/sys/unix.rs
```diff
@@ -659,6 +665,87 @@ impl SockAddr {
             None
         }
     }
+
+    /// Returns true if this address is an unnamed address from the `AF_UNIX` (AKA `AF_LOCAL`)
```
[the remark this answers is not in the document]
t4lz: Done https://github.com/rust-lang/socket2/pull/403/commits/b4ac3070142f2448b23c84b00d16882630d82a62

src/sys/unix.rs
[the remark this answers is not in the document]
Thomasdezeeuw: Odd that `AF_UNIX` is defined for Fuchsia as the check passes.

Thomasdezeeuw (approved):
LGTM. Only question is Fuchsia, we might have to remove it for now (sorry for suggesting it in the first place).

src/sys/unix.rs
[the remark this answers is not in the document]
tamird: Yep, it's defined, but not implemented.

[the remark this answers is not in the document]
ghanan94: @tamird is right, Fuchsia defines `AF_UNIX` but we do not (and have no plans to) implement unix domain sockets. 

We have `AF_UNIX` defined because its found in [musl](https://cs.opensource.google/fuchsia/fuchsia/+/main:zircon/third_party/ulib/musl/;drc=58986a79f3173fa3bb383f336df3828ca1f3d22d) "an MIT-licensed implementation of the standard C library targetting the Linux syscall API" that Fuchsia uses. 

Note including the unix address family on Fuchsia seems like the best path for now (please leave a comment mentioning this context so future changes do not accidentally add it).

src/sys/unix.rs
```diff
@@ -659,6 +665,87 @@ impl SockAddr {
...
+        })
+    }
+
+    /// Returns this address as a slice of bytes representing an abstract address if it is an
+    /// `AF_UNIX` abstract address, otherwise returns `None`.
+    ///
+    /// Abstract addresses are a Linux extension, so this method returns `None` on all non-Linux
+    /// platforms.
+    pub fn as_abstract_namespace(&self) -> Option<&[u8]> {
+        #[cfg(any(target_os = "linux", target_os = "android", target_os = "fuchsia"))]
```
Thomasdezeeuw, line 739:
```suggestion
        // NOTE: although Fuchsia does define `AF_UNIX` it's not actually implemented.
        #[cfg(any(target_os = "linux", target_os = "android"))]
```

src/sys/unix.rs
```diff
@@ -659,6 +665,87 @@ impl SockAddr {
...
+        {
+            self.as_sockaddr_un().and_then(|storage| {
+                (self.len() > offset_of_path(storage) as u32 && storage.sun_path[0] == 0)
+                    .then(|| self.path_bytes(storage, true))
+            })
+        }
+        #[cfg(not(any(target_os = "linux", target_os = "android", target_os = "fuchsia")))]
```
Thomasdezeeuw, line 746:
```suggestion
        #[cfg(not(any(target_os = "linux", target_os = "android")))]
```

tests/socket.rs
```diff
@@ -144,17 +146,52 @@ fn socket_address_unix() {
...
+    {
+        assert!(addr.is_unnamed());
+        assert_eq!(addr.as_pathname(), None);
+        assert_eq!(addr.as_abstract_namespace(), None);
+    }
+}
+
+#[test]
+#[cfg(all(
+    any(target_os = "linux", target_os = "android", target_os = "fuchsia"),
```
Thomasdezeeuw, line 180:
```suggestion
    any(target_os = "linux", target_os = "android"),
```

src/sys/unix.rs
```diff
@@ -659,6 +665,87 @@ impl SockAddr {
...
+
+    /// Returns true if this address is an unnamed address from the `AF_UNIX` family (for local
+    /// interprocess communication), false otherwise.
+    pub fn is_unnamed(&self) -> bool {
+        self.as_sockaddr_un()
+            .map(|storage| {
+                self.len() == offset_of_path(storage) as u32
+                    // On some non-linux platforms a zeroed path is returned for unnamed.
+                    // Abstract addresses only exist on Linux.
+                    || (cfg!(not(any(target_os = "linux", target_os = "android", target_os = "fuchsia")))
```
Thomasdezeeuw, line 677:
```suggestion
                    // NOTE: although Fuchsia does define `AF_UNIX` it's not actually implemented.
                    || (cfg!(not(any(target_os = "linux", target_os = "android")))
```

Thomasdezeeuw (commented):
One final suggestion based on @tamird and @ghanan94's feedback.

src/sys/unix.rs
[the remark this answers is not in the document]
ghanan94: Thanks! mind linking this thread (https://github.com/rust-lang/socket2/pull/403/commits/e78bbd942af997b813ea6a860fa86993f13d05f1#r1123557978) in the comment to leave breadcrumbs?

Thomasdezeeuw:
Thanks @t4lz!

PR #381, opened 2023-01-16
Add support for QNX Neutrino to 0.4 branch

flba-eb:
Change is close to https://github.com/rust-lang/socket2/pull/380 but not exactly the same.

CC @gh-tr

PR #419, opened 2023-03-24
Add support for 'all' feature on QNX Neutrino for v0.4.x

Tom-Goring:
Hi there,

I saw as part of #381 it was mentioned that the v0.4.x branch is not taking PRs, so apologies if this is not appropriate. When those changes were backported these `target_os` flags were not applied (I presume as the function signatures are totally different in master) which prevents compilation on 0.4 branches.

Not sure what the best way to resolve is - recommendations welcome :smile:

CC: @flba-eb @gh-tr

### DIFF
--- a/src/sockaddr.rs
+++ b/src/sockaddr.rs
@@ -234,7 +234,8 @@ impl From<SocketAddrV4> for SockAddr {
                 target_os = "ios",
                 target_os = "macos",
                 target_os = "netbsd",
-                target_os = "openbsd"
+                target_os = "openbsd",
+                target_os = "nto"
             ))]
             sin_len: 0,
         };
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw (approved):
> I saw as part of #381 it was mentioned that the v0.4.x branch is not taking PRs, so apologies if this is not appropriate. When those changes were backported these `target_os` flags were not applied (I presume as the function signatures are totally different in master) which prevents compilation on 0.4 branches.

Only backports are accept to make sure we don't miss any commits in newer versions, but this is a v0.4 specific fix, so it's fine.

Thomasdezeeuw:
Thanks @Tom-Goring

Tom-Goring:
Thanks very much!

PR #427, opened 2023-04-06
Add Protocol::UDPLITE

### LINKED ISSUE #361: UDPLITE support
Somewhat related to #350 is UDPLITE. It's supported on many an OS (linux and bsd for sure), and by default behaves just like UDP, with the exception that you can shorten the checksum, which is increasingly not terribly useful for its intended purpose, but it's there...

I'm getting kind of concerned  about UDP getting so heavily used by QUIC, limited port space from CGNATs, etc, and it's always seemed to me that at least trying to make udplite work was rather straightforward, and only a couple lines of code.

https://en.wikipedia.org/wiki/UDP-Lite

Thomasdezeeuw:
Closes #361

/cc @dtaht

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -318,9 +318,21 @@ impl Protocol {
     #[cfg_attr(docsrs, doc(cfg(all(feature = "all", target_os = "linux"))))]
     pub const DCCP: Protocol = Protocol(sys::IPPROTO_DCCP);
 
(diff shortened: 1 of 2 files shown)


Darksonn (approved):

Thomasdezeeuw:
Not sure what is wrong with Cirrus/FreeBSD's curl at the moment, but locally `cargo check` passes, so I'm going to merge.

PR #430, opened 2023-04-10
Fix Cirrus CI

Thomasdezeeuw:


Darksonn (approved):

Darksonn (commented):
The image already has curl, so you can also just remove the `pkg install` call.

Issue #433, opened 2023-04-15
error[E0308]: mismatched types when building for android


ShootingKing-AM:
- Succeeds with Building arm64-v8a (aarch64-linux-android) and Building x86_64 (x86_64-linux-android)
- Building x86 (i686-linux-android)
```rust
   --> C:\Users\user\.cargo\registry\src\index.crates.io-6f17d22bba15001f\socket2-0.5.2\src\sys\unix.rs:694:31
    |
694 |                 self.len() == offset_of_path(storage) as u32
    |                 ----------    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `i32`, found `u32`
    |                 |
    |                 expected because this is `i32`
    |
help: you can convert a `u32` to an `i32` and panic if the converted value doesn't fit
    |
694 |                 self.len() == (offset_of_path(storage) as u32).try_into().unwrap()
    |                               +                              +++++++++++++++++++++

error[E0308]: mismatched types
   --> C:\Users\user\.cargo\registry\src\index.crates.io-6f17d22bba15001f\socket2-0.5.2\src\sys\unix.rs:757:27
    |
757 |             (self.len() > offset_of_path(storage) as u32 && storage.sun_path[0] != 0).then(|| {
    |              ----------   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `i32`, found `u32`
    |              |
    |              expected because this is `i32`
    |
help: you can convert a `u32` to an `i32` and panic if the converted value doesn't fit
    |
757 |             (self.len() > (offset_of_path(storage) as u32).try_into().unwrap() && storage.sun_path[0] != 0).then(|| {
    |                           +                              +++++++++++++++++++++

error[E0308]: mismatched types
   --> C:\Users\user\.cargo\registry\src\index.crates.io-6f17d22bba15001f\socket2-0.5.2\src\sys\unix.rs:775:31
    |
775 |                 (self.len() > offset_of_path(storage) as u32 && storage.sun_path[0] == 0)
    |                  ----------   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `i32`, found `u32`
    |                  |
    |                  expected because this is `i32`
    |
help: you can convert a `u32` to an `i32` and panic if the converted value doesn't fit
    |
775 |                 (self.len() > (offset_of_path(storage) as u32).try_into().unwrap() && storage.sun_path[0] == 0)
    |                               +                              +++++++++++++++++++++

```

[...]

PR #434, opened 2023-04-16
Add support to remaining android targets

ShootingKing-AM:
fixes https://github.com/rust-lang/socket2/issues/433

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ include       = [
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
-targets = ["aarch64-apple-ios", "aarch64-linux-android", "x86_64-apple-darwin", "x86_64-unknown-fuchsia", "x86_64-pc-windows-msvc", "x86_64-pc-solaris", "x86_64-unknown-freebsd", "x86_64-unknown-illumos", "x86_64-unknown-linux-gnu", "x86_64-unknown-linux-musl", "x86_64-unknown-netbsd", "x86_64-unknown-redox"]
+targets = ["aarch64-apple-ios", "aarch64-linux-android", "x86_64-apple-darwin", "x86_64-unknown-fuchsia", "x86_64-pc-windows-msvc", "x86_64-pc-solaris", "x86_64-unknown-freebsd", "x86_64-unknown-illumos", "x86_64-unknown-linux-gnu", "x86_64-unknown-linux-musl", "x86_64-unknown-netbsd", "x86_64-unknown-redox", "armv7-linux-androideabi", "i686-linux-android"]
 
 [package.metadata.playground]
 features = ["all"]
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -691,7 +691,7 @@ impl SockAddr {
     pub fn is_unnamed(&self) -> bool {
         self.as_sockaddr_un()
             .map(|storage| {
-                self.len() == offset_of_path(storage).try_into().unwrap()
+                self.len() == offset_of_path(storage) as _
                     // On some non-linux platforms a zeroed path is returned for unnamed.
                     // Abstract addresses only exist on Linux.
                     // NOTE: although Fuchsia does define `AF_UNIX` it's not actually implemented.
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -29,7 +29,7 @@ include       = [
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
-targets = ["aarch64-apple-ios", "aarch64-linux-android", "x86_64-apple-darwin", "x86_64-unknown-fuchsia", "x86_64-pc-windows-msvc", "x86_64-pc-solaris", "x86_64-unknown-freebsd", "x86_64-unknown-illumos", "x86_64-unknown-linux-gnu", "x86_64-unknown-linux-musl", "x86_64-unknown-netbsd", "x86_64-unknown-redox"]
+targets = ["aarch64-apple-ios", "aarch64-linux-android", "x86_64-apple-darwin", "x86_64-unknown-fuchsia", "x86_64-pc-windows-msvc", "x86_64-pc-solaris", "x86_64-unknown-freebsd", "x86_64-unknown-illumos", "x86_64-unknown-linux-gnu", "x86_64-unknown-linux-musl", "x86_64-unknown-netbsd", "x86_64-unknown-redox", "armv7-linux-androideabi", "i686-linux-android"]
```
Thomasdezeeuw, line 32: Can you add to the `Makefile` as well.

src/sys/unix.rs
```diff
@@ -691,7 +691,7 @@ impl SockAddr {
     pub fn is_unnamed(&self) -> bool {
         self.as_sockaddr_un()
             .map(|storage| {
-                self.len() == offset_of_path(storage) as u32
+                self.len() == offset_of_path(storage).try_into().unwrap()
```
Thomasdezeeuw, line 694: Please use `as _` instead. I don't want to introduce a panic branch.

Thomasdezeeuw: Same for the other changes.

ShootingKing-AM: but, if `as _` fails, wont it make even harder to debug if type conversion is somehow messed up ?

Thomasdezeeuw: `offset_of_path` max is 108 bytes (off the top of my head), so it won't ever fail.

ShootingKing-AM: then it would not panic too, right ? why not try_into() ? This i guess is a much cleaner way, other way being conditional compilation using `#[cfg(...)]`

Thomasdezeeuw: > then it would not panic too, right ?

No, why would it panic?

> why not try_into() ?

As I said above that introduces a panic code branch, something we don't want.

> This i guess is a much cleaner way, other way being conditional compilation using `#[cfg(...)]`

Just use `as _`. The existing code was fine, Android just uses a different type `i32` instead of `u32`, both can easily fit the number 108.

ShootingKing-AM: I still don't get it (maybe nub at rust).
Please do push the required changes onto this PR

Thomasdezeeuw, line 694:
```suggestion
                self.len() == offset_of_path(storage) as _
```

ShootingKing-AM: Wow, didn't know that we can use `_` with `as`. Thought you were suggesting me to fill-in-the-blank after `as u32/i32`. Thanks learned something new today.

Don't want to sound argumentative but isn't it safer to use try_into() when we have no control over upstream crate (here `libc` which provides with `socklen_t`). Even if its a policy, why no panic blocks, is it just code optimization or there is something else deeper here ?

Thomasdezeeuw:
Thanks @ShootingKing-AM

src/sys/unix.rs
ShootingKing-AM: Understood. Thanks for your patience 🙏

dignifiedquire:
@Thomasdezeeuw any chance for a release that includes this soon? This came very timely for my project, as I need to compile for those androids.

Thomasdezeeuw:
@dignifiedquire sure, https://github.com/rust-lang/socket2/pull/440

PR #439, opened 2023-05-04
Haiku: update platform support.

jessicah:


jessicah:
I can do PRs for both? Targeted 0.4.x as that's what I need.

Thomasdezeeuw:
Only changes to master branch are accept, but I can backport it to v0.4 no problem. If we start accepting changes to v0.4.x as well it would be to easy to add a commit to v0.4.x that doesn't make it's way into v0.5.x.

PR #444, opened 2023-06-01
Add support for receiving ancillary data

yuguorui:
Two new methods are added in order to receive ancillary data along with normal data. The methods are `recv_with_ancillary_data` and `recv_vectored_with_ancillary_data_and_flags`. The former is easier to use, while the latter provides the flexibility.

The implementation includes changes to the `recvmsg` function in the `sys/unix.rs` file to handle the ancillary data.

A new test case is also added to test the `recv_with_ancillary_data` method.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -442,6 +442,29 @@ impl Socket {
         sys::recv(self.as_raw(), buf, flags)
     }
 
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw:
I assume this is continued in https://github.com/rust-lang/socket2/pull/446 and https://github.com/rust-lang/socket2/issues/445?

PR #446, opened 2023-06-02
[RFC] Add support for receiving ancillary data

yuguorui:
Two new methods are added in order to receive ancillary data along with normal data. The methods are `recv_with_ancillary_data` and `recv_vectored_with_ancillary_data_and_flags`. The former is easier to use, while the latter provides the flexibility.

The implementation includes changes to the `recvmsg` function in the `sys/unix.rs` file to handle the ancillary data.

A new test case is also added to test the `recv_with_ancillary_data` method.

Known issue: windows support is broken.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -950,7 +950,18 @@ pub(crate) fn recv_vectored(
     bufs: &mut [crate::MaybeUninitSlice<'_>],
     flags: c_int,
 ) -> io::Result<(usize, RecvFlags)> {
-    recvmsg(fd, ptr::null_mut(), bufs, flags).map(|(n, _, recv_flags)| (n, recv_flags))
+    recvmsg(fd, ptr::null_mut(), bufs, &mut [], flags).map(|(n, _, recv_flags)| (n, recv_flags))
+}
+
+#[cfg(not(target_os = "redox"))]
+pub(crate) fn recv_vectored_with_ancillary_data(
+    fd: Socket,
+    bufs: &mut [crate::MaybeUninitSlice<'_>],
+    ancillary_data: &mut [MaybeUninit<u8>],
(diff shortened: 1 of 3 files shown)


src/sys/unix.rs
```diff
@@ -950,7 +950,18 @@ pub(crate) fn recv_vectored(
     bufs: &mut [crate::MaybeUninitSlice<'_>],
     flags: c_int,
 ) -> io::Result<(usize, RecvFlags)> {
-    recvmsg(fd, ptr::null_mut(), bufs, flags).map(|(n, _, recv_flags)| (n, recv_flags))
+    recvmsg(fd, ptr::null_mut(), bufs, &mut [], flags).map(|(n, _, recv_flags)| (n, recv_flags))
```
Thomasdezeeuw, line 953: I think the addition of `&mut []` will be invalid as the pointer will be non-null.

Thomasdezeeuw (commented):
Maybe we should find a nice way to expose `recvmsg` (and `sendmsg`) instead of adding more methods that wrap it?

We'll need a nice wrapper around the `msghdr`/`WSAMSG` types. I might have some time to look into this this weekend.

Thomasdezeeuw:
@yuguorui what do you think about https://github.com/rust-lang/socket2/pull/447? Would that solve your use case (when I add `Socket::recvmsg`).

yuguorui:
> @yuguorui what do you think about #447? Would that solve your use case (when I add `Socket::recvmsg`).

Yes, I think it would be a great replacement for this PR. Thank you for your excellent work!

Thomasdezeeuw:
@yuguorui I think this can be closed now that we support `Socket::recvmsg` (https://github.com/rust-lang/socket2/pull/447)?

PR #463, opened 2023-08-27
windows-sys: use ADDRESS_FAMILY for `sa_family_t`

### LINKED ISSUE #462: Windows: `sa_family_t` should use `ADDRESS_FAMILY` from `windows-sys`
in #414 `socklen_t` and `sa_family_t` were changed to use types from `windows-sys`, which themselves were added to the `win32metadata` by request, in microsoft/win32metadata#1041 (both as `NativeTypedef`s). however, `sa_family_t` was later changed to `MetadataTypedef` in microsoft/win32metadata#1538 because it was an ["invented" handle type](https://github.com/microsoft/win32metadata/issues/1533#issuecomment-1507665486;) (meaning it doesn't exist in any of the Windows SDK header files but only in win32metadata). `socklen_t`, on the other hand, does exists in the Windows SDK header "WS2tcpip.h", so it remains `NativeTypedef`. 

it turns out that the equivalent to POSIX `sa_family_t` in Winsock2 is the `ADDRESS_FAMILY` type, and it existed forever. I think it's a mistake to request a `win32metadata` only typedef while ingoring the proper Windows SDK type,- and I'd suggest we change our `sa_family_t` to use the proper type, i.e. `windows_sys::Win32::Networking::Winsock::ADDRESS_FAMILY`, which also makes it consistent with other Winsock2 APIs, for instance, in `windows-sys` (and also `windows`), the type of `SOCKADDR::sa_family` and `SOCKADDR_STORAGE::sa_family` is `ADDRESS_FAMILY`, the constants `AF_INET`, `AF_INET6` are defined using type `ADDRESS_FAMILY`, and some iphelper functions like `GetIpForwardTable2()` accept `ADDRESS_FAMILY` as argument.

<details>
<summary>example code in the `windows` crate where `ADDRESS_FAMILY` is used</summary>

`SOCKADDR`

https://github.com/microsoft/windows-rs/blob/891c469706d314fcf0e3a7913f45970ce11b13f0/crates/libs/windows/src/Windows/Win32/Networking/WinSock/mod.rs#L13218-L13223

`SOCKADDR_IN::sin_family`

https://github.com/microsoft/windows-rs/blob/891c469706d314fcf0e3a7913f45970ce11b13f0/crates/libs/windows/src/Windows/Win32/Networking/WinSock/mod.rs#L13314-L13321

`AF_INET` and `AF_INET6`

https://github.com/microsoft/windows-rs/blob/891c469706d314fcf0e3a7913f45970ce11b13f0/crates/libs/windows/src/Windows/Win32/Networking/WinSock/mod.rs#L1782-L1785

</details>

btw, I discovered #414 when I was investigating the discrepancy of wrapper types from the `windows` crate. microsoft/windows-rs#2627

nerditation:
originally Windows SDK didn't have a `sa_family_t` typedef, and we requested `win32metadata` to add it (microsoft/win32metadata#1538) and used that from `windows-sys` as our `sa_family_t` definition on Windows platform (#414), which was an oversight. after all, Winsock2 isn't compliant to Posix and has it's own naming scheme, and `ADDRESS_FAMILY` is intended to be the equivalence of `sa_family_t`.

fixes #462

### DIFF
--- a/src/sys/windows.rs
+++ b/src/sys/windows.rs
@@ -61,7 +61,7 @@ pub(crate) use windows_sys::Win32::Networking::WinSock::{
     SOCKADDR_STORAGE as sockaddr_storage,
 };
 #[allow(non_camel_case_types)]
-pub(crate) type sa_family_t = windows_sys::Win32::Networking::WinSock::sa_family_t;
+pub(crate) type sa_family_t = windows_sys::Win32::Networking::WinSock::ADDRESS_FAMILY;
 #[allow(non_camel_case_types)]
 pub(crate) type socklen_t = windows_sys::Win32::Networking::WinSock::socklen_t;
 // Used in `Socket`.


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @nerditation

PR #469, opened 2023-09-11
Assign ss_len if it exists.

### LINKED ISSUE #468: Set `ss_len` when creating `SockAddr` from std
The `sockaddr_storage` type of BSD contains a field `ss_len` that represents the size of the sockaddr. When converting from `SocketAddr` etc., we need to set them correctly. Otherwise, it may report wrong result when comparing addresses.

Berrysoft:
Fixes #468 

I don't remove `len` field because:

* `socklen_t` is larger than `u8`.
* It will introduce more platform-specific code.
* Usual socket methods (at least on macOS) don't check `ss_len`. We still need another field to pass in and out.

### DIFF
--- a/src/sockaddr.rs
+++ b/src/sockaddr.rs
@@ -285,6 +285,23 @@ impl From<SocketAddrV4> for SockAddr {
             storage.sin_zero = Default::default();
             mem::size_of::<sockaddr_in>() as socklen_t
         };
+        #[cfg(any(
+            target_os = "dragonfly",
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw (approved):
I don't really know if we need it in another places as well. For example in calls to `accept`? Does the OS set `ss_len`?

Berrysoft:
AFAIK, macos sets ss_len to the correct value.

PR #474, opened 2023-09-21
Add GNU/Hurd support

sthibaul:


Thomasdezeeuw:
Can we add some CI support for this?  Running `cargo test` or `cargo check`?

sthibaul:
(a CI would need to wait for https://github.com/rust-lang/libc/pull/3325)

sthibaul:
Mmm, but AIUI, the current CI infrastructure installs support for the target from rustup? This is a tier3, so IIUC we won't have this?

Thomasdezeeuw:
> Mmm, but AIUI, the current CI infrastructure installs support for the target from rustup? This is a tier3, so IIUC we won't have this?

Maybe once https://github.com/rust-lang/socket2/pull/465 is merged we can add it to the tier 3 check, see https://github.com/rust-lang/socket2/pull/465/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R68.

sthibaul:
Ah, that'll have to wait for https://github.com/rust-lang/rust/pull/116527 apparently

Thomasdezeeuw:
Since https://github.com/rust-lang/rust/pull/116527 is merged, let's wait two days and restart the CI.

sthibaul:
Uh, between my tests and now, the use of `msg_iov` was added, and now I realize that the type was incorrect in libc, thus have to fix it there, and wait for another cycle.

sthibaul:
Ok, there it is, now it's passing :D

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @sthibaul

PR #481, opened 2023-11-21
haiku: Fix consumers by removing unsupported TOS calls

kallisti5:


kallisti5:
doing this in hopes of an eventual 0.4.11 :-)     A lot of stuff depends on socket2 0.4.10 which breaks a lot of crates on Haiku

kallisti5:
note: Targeting 0.4.x here.  If there are no plans for any more 0.4.x version releases, please ignore and close.

Thomasdezeeuw:
Is this already in v0.5? If not please target that and I'll backport.

kallisti5:
Ok.  These are all in 0.5.  :+1:

Thomasdezeeuw:
It's fine to go in to v0.4, I just want to make sure we're not losing attributes going from v0.4 to v0.5.

PR #485, opened 2023-12-18
Socket option for IP_MULTICAST_ALL & IPV6_MULTICAST_ALL for linux

mzachar:
Added socket options for IP_MULTICAST_ALL & IPV6_MULTICAST_ALL which can be used to filter multicast packets received on the socket only from the groups that have been explicitly joined.

Useful when used in combination with SO_REUSEPORT

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -1377,6 +1377,43 @@ impl Socket {
         }
     }
 
+    /// Get the value of the `IP_MULTICAST_ALL` option for this socket.
+    ///
+    /// For more information about this option, see [`set_multicast_all_v4`].
+    ///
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
Thanks @mzachar

Thomasdezeeuw:
Shit, actually this is missing a test. @mzachar can you add something like the following: https://github.com/rust-lang/socket2/blob/master/tests/socket.rs#L1322?

mzachar:
Certainly! Is it ok If I push new commit here?

Thomasdezeeuw:
> Certainly! Is it ok If I push new commit here?

I think you'll need to open a new pr.

PR #494, opened 2024-02-08
Add support for sendmmsg/recvmmsg

Tuetuopay:
This PR aims to add support for the `sendmmsg`/`recvmmsg` syscalls.

The API exposed here is a bit simplified, where the "user friendly" functions don't expose the scatter/gather stuff (the `_vectored` variants found for `recvmsg`/`sendmsg`). This was not exposed to not blow up the API surface, while keeping the functionality using `MmsgHdr(Mut)` for advanced usages.

Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -727,3 +727,151 @@ impl<'name, 'bufs, 'control> fmt::Debug for MsgHdrMut<'name, 'bufs, 'control> {
         "MsgHdrMut".fmt(fmt)
     }
 }
+
+/// Configuration of a `sendmmsg(2)` system call.
+///
+/// This wraps `mmsghdr` on Unix. Also see [`MmsgHdrMut`] for the variant used
+/// by `recvmmsg(2)`.
+#[cfg(not(target_os = "redox"))]
+pub struct MmsgHdr<'addr, 'bufs, 'control> {
+    inner: Vec<sys::mmsghdr>,
+    #[allow(clippy::type_complexity)]
+    _lifetimes: PhantomData<(&'addr SockAddr, &'bufs IoSlice<'bufs>, &'control [u8])>,
+}
+
+#[cfg(not(target_os = "redox"))]
+impl<'addr, 'bufs, 'control> MmsgHdr<'addr, 'bufs, 'control> {
+    /// Create a new `MmsgHdr` with all empty/zero fields.
+    #[allow(clippy::new_without_default)]
+    pub fn new(len: usize) -> MmsgHdr<'addr, 'bufs, 'control> {
+        // SAFETY: all zero is valid for `mmsghdr`
+        MmsgHdr {
+            inner: vec![unsafe { mem::zeroed() }; len],
+            _lifetimes: PhantomData,
+        }
+    }
+
+    /// Set the addresses (name) of the message.
+    ///
+    /// Corresponds to setting `msg_name` and `msg_namelen`.
+    pub fn with_addrs(mut self, addrs: &'addr [SockAddr]) -> Self {
+        for (msg, addr) in self.inner.iter_mut().zip(addrs) {
+            sys::set_msghdr_name(&mut msg.msg_hdr, addr);
+        }
+        self
+    }
+
+    /// Set the buffer(s) of the message.
+    ///
+    /// Corresponds to setting `msg_iov` and `msg_iovlen` on Unix.
+    pub fn with_buffers(mut self, bufs: &'bufs [IoSlice<'_>]) -> Self {
+        for (msg, buf) in self.inner.iter_mut().zip(bufs) {
+            sys::set_msghdr_iov(&mut msg.msg_hdr, buf as *const _ as *mut libc::iovec, 1);
+        }
+        self
+    }
+
+    /// Set the control buffer of the messages.
+    ///
+    /// Corresponds to setting `msg_control` and `msg_controllen` on Unix.
+    pub fn with_control(mut self, bufs: &'control [&'control [u8]]) -> Self {
+        for (msg, buf) in self.inner.iter_mut().zip(bufs) {
+            sys::set_msghdr_control(&mut msg.msg_hdr, buf.as_ptr() as *mut _, buf.len())
+        }
+        self
+    }
+
+    /// Set the flags on the messages
+    ///
+    /// Corresponds to setting `msg_flags` on Unix.
+    pub fn with_flags(mut self, flags: &[sys::c_int]) -> Self {
+        for (msg, flags) in self.inner.iter_mut().zip(flags) {
+            sys::set_msghdr_flags(&mut msg.msg_hdr, *flags);
+        }
+        self
+    }
+}
+
+#[cfg(not(target_os = "redox"))]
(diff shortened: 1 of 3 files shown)


src/sys/unix.rs
```diff
@@ -1067,6 +1085,97 @@ pub(crate) fn recvmsg(
...
+        {
+            #[cfg(target_os = "freebsd")]
+            {
+                msgs.inner.len()
+            }
+            #[cfg(not(target_os = "freebsd"))]
+            {
+                msgs.inner.len() as u32
+            }
+        },
```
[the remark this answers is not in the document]
Tuetuopay: gah i'm an idiot for not thinking about this one. thanks!

src/lib.rs
```diff
@@ -727,3 +727,211 @@ impl<'name, 'bufs, 'control> fmt::Debug for MsgHdrMut<'name, 'bufs, 'control> {
...
+    not(any(
+        target_os = "redox",
+        target_os = "solaris",
+        target_os = "hurd",
+        target_os = "vita",
+        target_os = "illumos",
+        target_vendor = "apple"
+    ))
+))]
+pub struct MmsgHdr<'addr, 'bufs, 'control> {
```
Thomasdezeeuw, line 746: What about matching `mmsghdr` here? For example 

```rust
pub struct MmsgHdr<'addr, 'bufs, 'control> {
    msg: MsgHdr,
    len: libc::c_uint,
}
```

This way the caller can decide if they want to use a `Vec`, a slice an array or something else.

xv-ian-c:
@Tuetuopay are you still interested in pursuing this functionality?

src/lib.rs
```diff
@@ -727,3 +727,211 @@ impl<'name, 'bufs, 'control> fmt::Debug for MsgHdrMut<'name, 'bufs, 'control> {
...
+    /// `Control` on Windows.
+    pub fn with_control(mut self, buf: &'control mut [&'control mut [MaybeUninit<u8>]]) -> Self {
+        for (msg, buf) in self.inner.iter_mut().zip(buf) {
+            sys::set_msghdr_control(&mut msg.msg_hdr, buf.as_mut_ptr().cast(), buf.len());
+        }
+        self
+    }
+
+    /// Returns the flags of the message.
+    pub fn flags(&self, n: usize) -> Vec<RecvFlags> {
```
xv-ian-c, line 913: I think accessors are also needed for `len` and `control_len` fields in order to be able to usefully consume the result of the syscall.

Rather than the 3 vec allocations it might also be nice to have a method to return all 3 at once so we'd at least have only 1, or to avoid it completely with an iterator based API.

Tuetuopay:
> @Tuetuopay are you still interested in pursuing this functionality?

@xv-ian-c Hi! Thanks for the reminder. I completely forgot about this. Yes I still mean to get this through, whenever I get time to work on this :)

src/lib.rs
xv-ian-c:
```diff
diff --git a/src/lib.rs b/src/lib.rs
index 6c49cd1..51ebcc2 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -923,7 +923,43 @@ impl<'addr, 'bufs, 'control> MmsgHdrMut<'addr, 'bufs, 'control> {
         self
     }
 
-    /// Returns the flags of the message.
+    /// Returns the lengths, flags and control lengths of the messages.
+    pub fn results(&self, n: usize) -> Vec<(usize, RecvFlags, usize)> {
+        self.inner
+            .iter()
+            .take(n)
+            .map(|msg| {
+                (
+                    msg.msg_len as _,
+                    sys::msghdr_flags(&msg.msg_hdr),
+                    sys::msghdr_control_len(&msg.msg_hdr),
+                )
+            })
+            .collect()
+    }
+
+    /// Extends the vec with the length, flags and control lengths of the messages.
+    /// This avoids the need to allocate a new vec on each use which affects `results`.
+    pub fn extend_with_results(&self, v: &mut Vec<(usize, RecvFlags, usize)>, n: usize) {
+        v.extend(self.inner.iter().take(n).map(|msg| {
+            (
+                msg.msg_len as _,
+                sys::msghdr_flags(&msg.msg_hdr),
+                sys::msghdr_control_len(&msg.msg_hdr),
+            )
+        }));
+    }
+
+    /// Returns the lengths of the messages.
+    pub fn lens(&self, n: usize) -> Vec<usize> {
+        self.inner
+            .iter()
+            .take(n)
+            .map(|msg| msg.msg_len as _)
+            .collect()
+    }
+
+    /// Returns the flags of the messages.
     pub fn flags(&self, n: usize) -> Vec<RecvFlags> {
         self.inner
             .iter()
@@ -931,6 +967,15 @@ impl<'addr, 'bufs, 'control> MmsgHdrMut<'addr, 'bufs, 'control> {
             .map(|msg| sys::msghdr_flags(&msg.msg_hdr))
             .collect()
     }
+
+    /// Returns the control lengths of the messages.
+    pub fn control_lens(&self, n: usize) -> Vec<usize> {
+        self.inner
+            .iter()
+            .take(n)
+            .map(|msg| sys::msghdr_control_len(&msg.msg_hdr))
+            .collect()
+    }
 }
 
 #[cfg(all(
```

Is what I ended up with.

Thomasdezeeuw (changes requested):
I'm not going to accept all the usage of `Vec`tors here, we can do this without allocations.

src/lib.rs
Tuetuopay: I quite like the idea of an iterator-based API. This lets the user choose if they want a Vec.

src/lib.rs
Tuetuopay: I see two options:

- go with your suggestion, and let callers provide their own slices of MmsgHdr. This works for the "lower level" `Socket::sendmmsg` function, but not for `Socket::send_multiple_to` as it needs some memory layout adaptation
- const generics may be an option? though we may still want to defer to the caller for that

src/lib.rs
xv-ian-c: It's not too bad for `results`, `lens` and `control_lens` since they are `Copy` but for `recvmmsg` getting back mutable access to the received buffer is a bit of a lifetime quagmire.

You basically setup the largest number of buffers you want to receive at once, which is fine if you use them all, but often you'll receive some smaller number.

The overhead of repopulating from scratch on each iteration is noticeable enough that it leads to populating a smaller number of potential receives than you otherwise would want to.

Ideally you'd want to repopulate the buffers which did get used but reuse the ones which didn't. I've absolutely failed to make that work with the lifetimes involved, since the `MmsgHdr`s end up holding a mutable reference into whatever owns the destination buffers[^1], which means one needs to drop the `mmsghdr` which you might otherwise be able to partially reuse.

AFAICT the same lifetime issue exists whether the collection in `MsgHdrMut` is internal (as in the `Vec` here) or external (as in a slice of the `MsgHdrMut` proposed in https://github.com/rust-lang/socket2/pull/494#discussion_r1498132735).

Is there some technique where the lifetime of the (used) head can be split from the (unused) tail?

[^1]: In my case the buffers are owned by an array of BytesMut, but it could be a `Vec<Vec<u8>>`  or whatever.

Thomasdezeeuw:
Closing due to age.

PR #499, opened 2024-03-12
Add MSG_CONFIRM and MSG_DONTROUTE to RecvFlags

pd0wm:
This is useful when reading from SocketCAN sockets. The MSG_CONFIRM flag is used to indicate a successful transmission of a CAN frame. 

Reference: https://www.kernel.org/doc/html/next/networking/can.html#raw-socket-option-can-raw-join-filters

Is there any reason the raw flags value is not exposed to the user? There might be more drivers that are using non-standard flags to communicate a certain state.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -573,6 +573,37 @@ impl RecvFlags {
     pub const fn is_out_of_band(self) -> bool {
         self.0 & libc::MSG_OOB != 0
     }
+
+    /// Check if the confirm flag is set.
+    ///
+    /// This is used by SocketCAN to indicate a frame was sent via the
+    /// socket it is received on. This flag can be interpreted as a
+    /// 'transmission confirmation'.
+    ///
+    /// On Unix this corresponds to the `MSG_CONFIRM` flag.
+    #[cfg(all(feature = "all", any(target_os = "android", target_os = "linux")))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(feature = "all", any(target_os = "android", target_os = "linux"))))
+    )]
+    pub const fn is_confirm(self) -> bool {
+        self.0 & libc::MSG_CONFIRM != 0
+    }
+
+    /// Check if the don't route flag is set.
+    ///
+    /// This is used by SocketCAN to indicate a frame was created
+    /// on the local host.
+    ///
+    /// On Unix this corresponds to the `MSG_DONTROUTE` flag.
+    #[cfg(all(feature = "all", any(target_os = "android", target_os = "linux")))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(feature = "all", any(target_os = "android", target_os = "linux"))))
+    )]
+    pub const fn is_dontroute(self) -> bool {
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
> Is there any reason the raw flags value is not exposed to the user? There might be more drivers that are using non-standard flags to communicate a certain state.

We're trying to provide a cross platform API, but for Windows `RecvFlags` is completely fake.

pd0wm:
Looks like the build is failing due to missing fields for some targets. Shall I limit the new functions to `target_os = "linux"`?

Thomasdezeeuw:
> Looks like the build is failing due to missing fields for some targets. Shall I limit the new functions to `target_os = "linux"`?

It's it's Linux only then yes. But it will require the `all` feature as well.

pd0wm:
Should be supported on Android as well. Is this what you mean by requiring both the `all` feature and gating on the os?

src/sys/unix.rs
```diff
@@ -573,6 +573,37 @@ impl RecvFlags {
...
+    #[cfg(all(feature = "all", any(target_os = "android", target_os = "linux")))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(feature = "all", any(target_os = "android", target_os = "linux"))))
+    )]
+    pub const fn is_confirm(self) -> bool {
+        self.0 & libc::MSG_CONFIRM != 0
+    }
+
+    /// Check if the don't route flag is set
```
Thomasdezeeuw, line 593: Nit.
```suggestion
    /// Check if the don't route flag is set.
```

Thomasdezeeuw (commented):
Is there a way we can get tests for this? I know the code is fairly trivial, but would be nice to know it works.

Otherwise the code LGTM, minus some small docs nits.

pd0wm:
I think `MSG_CONFIRM` is normally used a a flag for calling `sendmsg` not on received messages. So not sure if I can trigger this with an end-to-end test without using SocketCAN. SocketCAN cannot be easily tested inside a docker container, as the host needs to create a virtual SocketCAN device first.

Thomasdezeeuw:
> Should I add the flags to the impl `std::fmt::Debug` as well? https://github.com/rust-lang/socket2/blob/master/src/sys/unix.rs#L579-L589
> 
> I can add a test that the value is false under normal circumstances here: https://github.com/rust-lang/socket2/blob/master/tests/socket.rs#L685-L686

Yes to both. After that we can merge.

pd0wm:
> > Should I add the flags to the impl `std::fmt::Debug` as well? https://github.com/rust-lang/socket2/blob/master/src/sys/unix.rs#L579-L589
> > I can add a test that the value is false under normal circumstances here: https://github.com/rust-lang/socket2/blob/master/tests/socket.rs#L685-L686
> 
> Yes to both. After that we can merge.

Already done :)

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @pd0wm

pd0wm:
Thanks for your quick feedback!

PR #505, opened 2024-04-15
Add `MsgHdrMut::control_len` to get how much of control buffer was filled.

tobz:
As stated in the title, this exposes the underlying `msg_controllen`/`Control.len` field for `MsgHdrMut`, which is necessary to figure out how many bytes, if any, were filled in the control buffer.


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @tobz

PR #507, opened 2024-04-22
Prevent MsgHdr::with_control with empty buffer causes error in macOS

youknowone:
Otherwise panic when `cfg(debug_assertions)` will be another good prevention.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -639,8 +639,11 @@ impl<'addr, 'bufs, 'control> MsgHdr<'addr, 'bufs, 'control> {
     /// Corresponds to setting `msg_control` and `msg_controllen` on Unix and
     /// `Control` on Windows.
     pub fn with_control(mut self, buf: &'control [u8]) -> Self {
-        let ptr = buf.as_ptr() as *mut _;
-        sys::set_msghdr_control(&mut self.inner, ptr, buf.len());
+        debug_assert!(
+            !buf.is_empty(),
+            "Empty control buffer might be a programming error"
+        );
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw (commented):
Are there valid reasons to send an empty control value? Or is this always programmer error?

youknowone:
To tell about the valid reasons, I am not familiar with low level socket API that much.
At least it was a programmer error for my case. Is panic better?

Thomasdezeeuw:
> To tell about the valid reasons, I am not familiar with low level socket API that much. At least it was a programmer error for my case. Is panic better?

No, I would say not doing anything would be better. We're not going to introduce panics in our API, but hiding errors/problematic code is also not something we want.

youknowone:
It was a small glitch but painful to debug.
I finally could find the buffer is wrong, but it was hidden inside socket2 crate.
As a library user, I could see something goes wrong on `sendmsg` but not on `with_control`. If `with_control` is not expected to be called with empty buffer by nature, it could be warned at the moment, instead of deferring it.
Maybe one of the reason was `Debug for MsgHdr` was useless and didn't display its contents.

I changed the patch to `debug_assert!()` make it neither panic(in real use) nor hiding and deferring error. Please take a review if you agree it is a UX caveat.

Thomasdezeeuw:
> I understand your concern. Is there any plan to implement better Debug for MagHdr in that case? Debug can be platform-specific implementation and it will not affect any actual behavior. Minor platforms still can use current implementation

A pr for that would be accepted.

Thomasdezeeuw:
Closing this since I still don't think it's something we want.

PR #510, opened 2024-04-28
Support transparent proxy related unix socket options as possible

cavivie:
Changes:
- (get/set) ip_transparent (v4) should be supported on Android in theory.

Features:
- new apis`ip_transparent_v6`/`set_ip_transparent_v6` to support ipv6 transparent on (linux/android).
- new apis `so_bindany`/`set_so_bindany` to support bind any address to make a transparent proxy on (openbsd).
- new apis `ip_bindany[_v6]`/`set_ip_bindany[_v6]` to support bind any address to make a transparent proxy on (freebsd).


cavivie:
Will this request cause any problems? If so, please let me know. Otherwise, should I close this PR? I don't want to keep it open for a long time.

Thomasdezeeuw:
Sorry I missed this pr, will try to review this later this week.

Thomasdezeeuw (commented):
Can you split this up into multiple commits?

* the move to `src/sys/unix.rs`
* adding target_os = "fuchsia"
* adding `(set_)ip_transparent_v6`
* adding `(set_)ip_bindany_v4` (instead of `ip_bindany`)
* adding `(set_)ip_bindany_v6`
* deprecate `ip_transparent` in favour of `ip_transparent_v4`, we're slowly moving to `_v4` and `_v6` versions

cavivie:
> Can you split this up into multiple commits?
> 
> 
> 
> * the move to `src/sys/unix.rs`
> 
> * adding target_os = "fuchsia"
> 
> * adding `(set_)ip_transparent_v6`
> 
> * adding `(set_)ip_bindany_v4` (instead of `ip_bindany`)
> 
> * adding `(set_)ip_bindany_v6`
> 
> * deprecate `ip_transparent` in favour of `ip_transparent_v4`, we're slowly moving to `_v4` and `_v6` versions

OK, do this work later.

cavivie:
@Thomasdezeeuw  Quick question, `SO_BINDANY` supports both v4 and v6 on OpenBSD, should we use (set_)so_bindany or (set_)so_bindany_(v4|v6)?

cavivie:
The last freebsd ci failed and I can't determine exactly why. The only thing that is different is that the old code was able to pass ci and the ci image version was `freebsd-13-2`, the new one seems to have changed to `freebsd-13-2-release-amd`.

Thomasdezeeuw (commented):
Thanks for splitting up the commit, makes it much easier to review this way.

I'm not familiar with these options myself. But looking at the docs it seems like these options do roughly the same thing, they are just set differently on different OS? Is that correct?

If so, I think we should try and merge the functions such that the user doesn't need a bunch of `cfg` to make the correct call depending on the OS.

cavivie:
As the title says, we should not merge these behaviors, some options themselves are not the meaning of transparent proxy, they are necessary settings to complete transparent proxy, but not the only settings.
Although their functionalities overlap somewhat, due to their different design purposes and use cases, they have distinct focuses and implementations in practice.

cavivie:
@Thomasdezeeuw Or we should merge (set_)freebind/(set_)_ip_bindany/(set_)_so_bindany, which should be equivalent options on different platforms. For Linux, we only need (set_)_ip_transparent. WDYT?

Thomasdezeeuw:
I prefer the second option with proper document as to what options are being set, iff the options are actually the same. We have something similar for the keep alive options.

cavivie:
Now coded according to the new requirements.

PR #513, opened 2024-05-24
Ignore clippy new_without_default

Thomasdezeeuw:
Not needed for TcpKeepalive.


Darksonn (approved):
You sure you don't want to just add a `Default` impl?

Thomasdezeeuw:
> You sure you don't want to just add a `Default` impl that calls `new`?

If someone needs it we can consider it, but for now it's just more API surface we would have to maintain.

Issue #487, opened 2024-01-08
Currently socket2 has no way to set the DF bit on packets. The attached patch adds it.


jcdutton:
[socket2.diff.txt](https://github.com/rust-lang/socket2/files/13859057/socket2.diff.txt)
Hi,
Currently socket2 has no way to set the DF bit on packets.
The DF bit is needed for Path MTU discovery and also for creating arbitrary ICMP packets with the DF bit set.
Please see the attached patch that will add it.
I have only been able to test it on Linux, not Windows.
This might also be needed to enable Path MTU discovery on TCP sessions when using socket2.

For example, some sample code to send an ICMP packet with the DF bit set:
    let domain = Domain::IPV4;
    let ty =  Type::DGRAM;
    let protocol = Some(Protocol::ICMPV4);
    let mut sock1 = create_socket(domain, ty, protocol);
    let mut sock1r = sock1.unwrap();
    let res1 = Socket::set_mtu_discover(&sock1r, true);
    let mut sock2 = sock1r;
    sock2.set_read_timeout(Some(timeout))?;
    let destination = SocketAddr::new(addr.clone(), 0); // destination
    let sockaddr1: SockAddr = destination.into();
    let result = make_data(&data, size3);
    let payload = result.unwrap();
    let res2 = sock2.send_to(&payload, &sockaddr1);

jcdutton:
@Thomasdezeeuw 
> Pull request would be welcome.
I would like someone else to check it does not break socket2 on other platforms first (e.g. windows), and maybe re-craft it to work on more platforms before we commit it. For example, I don't know if I have put "use libc;" in a place that will work on windows platforms.
I have tested it on Linux x86_64 but nothing else.

Thomasdezeeuw:
Closing as a duplicate of #318.

PR #515, opened 2024-06-09
#487 Add ways to set Path Discovery MTU for Linux on a Socket.

darnuria:
Add ways to set Path Discovery MTU for Linux on a Socket.

This contribution use setsocketops to configure Dont Fragment (DF) bit and OS
behavior related to PMTU.

It introduce PathMtuDiscoveringMode enum to manage it rust-way and 4 functions
respectively for ipv4/v6 and set/get.

Omit and Interface exist but are not widely documented.

Initially It started by preparing a PR for the patch purposed here: https://github.com/rust-lang/socket2/issues/487

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -1391,6 +1391,39 @@ pub(crate) const fn to_mreqn(
     }
 }
 
+/// Possibles state for Path Maximum Transmission Unit (PMTU) Discovering of packets received on a socket.
+///
+/// It set in the IP packet Header the flag DF also known as "do not fragment".
+///
+/// For UDP it's particullary important to note that during Path MTU discovery is done by the OS,
+/// incomming datagram may be dropped.
+///
+/// ## Portability
+///
+/// It's Linux only way of managing the Path MTU.
+#[cfg(all(feature = "all", target_os = "linux"))]
+#[cfg_attr(docsrs, doc(cfg(all(feature = "all", target_os = "linux"))))]
+#[repr(C)]
+#[derive(Debug)]
+pub enum PathMtuDiscoveringMode {
+    /// Path MTU discovery is not done.
+    Dont = 0,
+    /// Path MTU discovery is done according to route setting.
+    Want = 1,
+    /// Path MTU discovery
+    Do = 2,
+    /// Set DF bit but ignore Path MTDU
+    Probe = 3,
+    /// Use interface MTU.
+    /// It ignore destination PMTU and does not set DF flag.
+    /// Incomming ICMP frag_needed notifications on this socket will be ignored in
+    /// order to prevent accepting spoofed ones.
+    Interface = 4,
+    /// Almost like [MtuDiscoveringMode::Interface] but authorize fragmented packet
+    /// if they do not saturate the interface MTU.
+    Omit = 5,
+}
+
 /// Unix only API.
 impl crate::Socket {
     /// Accept a new incoming connection from this listener.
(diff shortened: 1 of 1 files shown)


src/sys/unix.rs
```diff
@@ -1391,6 +1391,39 @@ pub(crate) const fn to_mreqn(
...
+/// incomming datagram may be dropped.
+///
+/// ## Portability
+///
+/// It's Linux only way of managing the Path MTU.
+#[cfg(all(feature = "all", target_os = "linux"))]
+#[cfg_attr(docsrs, doc(cfg(all(feature = "all", target_os = "linux"))))]
+#[repr(C)]
+#[derive(Debug)]
+pub enum PathMtuDiscoveringMode {
```
Thomasdezeeuw, line 1408: I'm not sure about this enum.

Maybe it's better as a type alias with constants for the options? This way we don't have to do the conversion and we don't have to deal with the case where the kernel returns a value we don't support yet.

src/sys/unix.rs
```diff
@@ -1391,6 +1391,39 @@ pub(crate) const fn to_mreqn(
...
+/// Possibles state for Path Maximum Transmission Unit (PMTU) Discovering of packets received on a socket.
+///
+/// It set in the IP packet Header the flag DF also known as "do not fragment".
+///
+/// For UDP it's particullary important to note that during Path MTU discovery is done by the OS,
+/// incomming datagram may be dropped.
+///
+/// ## Portability
+///
+/// It's Linux only way of managing the Path MTU.
```
Thomasdezeeuw, line 1403: This is not needed as it's already shown in the docs, e.g. see https://docs.rs/socket2/latest/socket2/struct.Socket.html#method.pair.
```suggestion
```

darnuria:
Hello, thanks for the review, will do the changes today.

src/sys/unix.rs
darnuria: Ok I agree, I may do a patch to the linux man page since some are not explicitly documented outside of the `in.h`/`in6.h`

darnuria: > Maybe it's better as a type alias with constants for the options? This way we don't have to do the conversion and we don't have to deal with the case where the kernel returns a value we don't support yet.

I reused the pattern used for `Type` and `Protocol` code tell me if wasn't the best move. :)

src/sys/unix.rs
```diff
@@ -1391,6 +1391,134 @@ pub(crate) const fn to_mreqn(
...
+/// This type is freely interconvertible with C's `int` type, however, if a raw
+/// value needs to be provided.
+#[cfg(all(feature = "all", target_os = "linux"))]
+#[cfg_attr(docsrs, doc(cfg(all(feature = "all", target_os = "linux"))))]
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub struct PathMtuDiscoveringModeV6(libc::c_int);
+
+#[cfg(all(feature = "all", target_os = "linux"))]
+#[cfg_attr(docsrs, doc(cfg(all(feature = "all", target_os = "linux"))))]
+impl PathMtuDiscoveringModeV6 {
```
darnuria, line 1475: It involve some code duplication alas but it's code that... never change except for adding new mode since API don't change for Linux.

Thomasdezeeuw:
Sorry @darnuria I didn't have time to review this last weekend and I don't think I'll have time the next two weeks either, so it's going to take a little while.

larseggert:
It would be great if this PR would also allow setting `IP_DONTFRAG` for UDP sockets.

PR #520, opened 2024-06-19
Merge Check, CheckTier3, Docs and DocsTier3 CI jobs

Thomasdezeeuw:
And expand the amount of targets we check.

Instead of getting the target setup twice, one for the cargo check run and another for cargo doc, do it once and check both.


Thomasdezeeuw:
This basically copies Mio's check job to socket2.

Darksonn (approved):

PR #536, opened 2024-10-21
feat: Add UDP GRO option

XAMPPRocky:
Allows setting UDP_GRO on the socket.

https://www.man7.org/linux/man-pages/man7/udp.7.html

> Enables UDP receive offload.  If enabled, the socket may
              receive multiple datagrams worth of data as a single large
              buffer, together with a cmsg(3) that holds the segment
              size.  This option is the inverse of segmentation offload.
              It reduces receive cost by handling multiple datagrams
              worth of data as a single large packet in the kernel
              receive path, even when that exceeds MTU.  This option
              should not be used in code intended to be portable.

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -3097,6 +3097,36 @@ impl crate::Socket {
             )
         }
     }
+
+    /// Get the value of the `UDP_GRO` option on this socket.
+    ///
+    /// For more information about this option, see [`set_udp_gro`].
+    ///
+    /// [`set_udp_gro`]: Socket::set_udp_gro
+    #[cfg(all(feature = "all", any(target_os = "android", target_os = "linux")))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(feature = "all", any(target_os = "android", target_os = "linux"))))
+    )]
+    pub fn udp_gro(&self) -> io::Result<bool> {
+        unsafe {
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw (changes requested):
Since this Linux only, can you put in the sys/unix.rs file: https://github.com/rust-lang/socket2/blob/3a938932829ea6ee3025d2d7a86c7b095c76e6c3/src/sys/unix.rs#L763

Could also you make sure it's order by option name.

Can you add the appropriate attributes, e.g. see https://github.com/rust-lang/socket2/blob/3a938932829ea6ee3025d2d7a86c7b095c76e6c3/src/sys/unix.rs#L791-L795

Finally can you add a test? A simple set and get test is sufficient, see https://github.com/rust-lang/socket2/blob/3a938932829ea6ee3025d2d7a86c7b095c76e6c3/tests/socket.rs#L1357

XAMPPRocky:
@Thomasdezeeuw Thank you, I've implemented your feedback, the only thing I wasn't 100% sure on was "Could also you make sure it's order by option name." let me know if I need to make more changes

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @XAMPPRocky

XAMPPRocky:
Hey, yeah my initial need for the option isn't needed as the project I was using it in now uses XDP sockets which bypass this setting. I'lll try to mixup the PR next week though.

Thomasdezeeuw:
Closing this due to inactivity.

PR #556, opened 2025-02-20
feat: add dont_frag option

### LINKED ISSUE #318: support set IP_DONTFRAG
see

https://chromium.googlesource.com/chromium/src/net/+/refs/heads/main/socket/udp_socket_posix.cc
 ```
  | // setsockopt(IP_DONTFRAG) is supported on macOS from Big Sur
  | #elif BUILDFLAG(IS_MAC)
  | if (!base::mac::IsAtLeastOS11()) {
  | return ERR_NOT_IMPLEMENTED;
  | }
  | int val = 1;
  | if (addr_family_ == AF_INET6) {
  | int rv =
  | setsockopt(socket_, IPPROTO_IPV6, IPV6_DONTFRAG, &val, sizeof(val));
  | // IP_DONTFRAG is not supported on v4mapped addresses.
  | return rv == 0 ? OK : MapSystemError(errno);
  | }
  | int rv = setsockopt(socket_, IPPROTO_IP, IP_DONTFRAG, &val, sizeof(val));
  | return rv == 0 ? OK : MapSystemError(errno);
 ```

<br class="Apple-interchange-newline">

cathaysia:
fix #318

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -1615,6 +1615,53 @@ impl Socket {
                 .map(|recv_tos| recv_tos > 0)
         }
     }
+
+    /// Set value for `IP_DONTFRAG` option of this socket
+    #[cfg(any(
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "linux",
+        target_os = "windows",
+    ))]
+    pub fn dont_frag(&self) -> io::Result<bool> {
+        unsafe {
+            #[cfg(any(target_os = "macos", target_os = "ios",))]
+            return getsockopt::<c_int>(self.as_raw(), sys::IPPROTO_IP, libc::IP_DONTFRAG)
+                .map(|dont_frag| dont_frag > 0);
+            #[cfg(any(target_os = "linux", target_os = "windows",))]
+            return getsockopt::<c_int>(self.as_raw(), sys::IPPROTO_IP, sys::IP_MTU_DISCOVER)
+                .map(|dont_frag| dont_frag == 0);
+        }
+    }
+    /// Set value for `IP_DONTFRAG` option on this socket
+    #[cfg(any(
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "linux",
+        target_os = "windows",
+    ))]
+    pub fn set_dont_frag(&self, dont_frag: bool) -> io::Result<()> {
+        unsafe {
+            #[cfg(any(target_os = "macos", target_os = "ios",))]
+            return setsockopt(
+                self.as_raw(),
+                sys::IPPROTO_IP,
+                libc::IP_DONTFRAG,
+                dont_frag as c_int,
+            );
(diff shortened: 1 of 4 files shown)


cathaysia:
what happend about this ci?

src/lib.rs
```diff
@@ -381,7 +381,7 @@ impl RecvFlags {
 #[repr(transparent)]
 pub struct MaybeUninitSlice<'a>(sys::MaybeUninitSlice<'a>);
 
-impl<'a> fmt::Debug for MaybeUninitSlice<'a> {
```
Thomasdezeeuw, line 384: Can you revert these unrelated changes?

src/socket.rs
```diff
@@ -1615,6 +1615,101 @@ impl Socket {
...
+
+    /// Set value for `IP_DONTFRAG` option of this socket
+    #[cfg(not(any(
+        target_os = "netbsd",
+        target_os = "openbsd",
+        target_os = "redox",
+        target_os = "vita",
+        target_os = "dragonfly",
+        target_os = "fuchsia",
+    )))]
```
Thomasdezeeuw, line 1627: Can you do the inverse of this, that is specifying why OS do support this option.

src/sys/windows.rs
```diff
@@ -1022,36 +1035,41 @@ impl FromRawSocket for crate::Socket {
     }
 }
 
-#[test]
```
Thomasdezeeuw, line 1025: Why did you make these changes? Please revert.

src/sys/windows.rs
```diff
@@ -84,6 +84,19 @@ pub(crate) use windows_sys::Win32::Networking::WinSock::{
...
+    };
+    ($id:ident) => {
+        export_constants!($id, $id);
+    };
+}
+
+export_constants!(IPV6_MTU_DISCOVER);
+export_constants!(IP_MTU_DISCOVER);
+export_constants!(IP_PMTUDISC_DO);
+export_constants!(IP_PMTUDISC_DONT);
```
Thomasdezeeuw, line 99: Can you please remove this macro and use the same style we do for everything else?

Thomasdezeeuw (changes requested):
This now has way too many unrelated changes. Please don't overload a pr with unrelated stuff.

If you want to make multiple changes, please open multiple prs.

src/sys/windows.rs
cathaysia: if you don't warp #[test] in cfg(test), then these code will be compiled into the binary.

src/socket.rs
cathaysia: I don't know how other OS work. I changed the cfg to macos/ios/linux/windows. Other platforms need others to fill in

larseggert:
Is there an alternative PR? I would like to see this functionality added.

PR #557, opened 2025-02-21
Add tcp retries to windows

keithmattix:
Bump windows sys and add support for retries on Windows

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -441,7 +441,6 @@ pub struct TcpKeepalive {
         target_os = "openbsd",
         target_os = "redox",
         target_os = "solaris",
-        target_os = "windows",
         target_os = "nto",
         target_os = "espidf",
         target_os = "vita",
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -54,7 +55,7 @@ features = ["all"]
 libc = "0.2.150"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.52"
+version = "0.59"
```
Thomasdezeeuw, line 58: This is a breaking change, can't accept that at the moment, see #526

keithmattix: Do you need a volunteer to implement https://github.com/rust-lang/socket2/issues/526#issuecomment-2557143907? I'm happy to take a stab at it, or pass the feedback to someone internally at MSFT

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @keithmattix

PR #561, opened 2025-02-26
Enable `IP_BOUND_IF` on illumos and Solaris

### LINKED ISSUE #560: `Socket::bind_device_by_index_{v4,v6}` should be available on illumos/Solaris
The `Socket::bind_device_by_index_v4` and `Socket::bind_device_by_index_v6` functions wrap the socket option `IP_BOUND_IF`. Presently, these functions have the following `cfg` attribute:

```rust
    #[cfg(all(
        feature = "all",
        any(
            target_os = "ios",
            target_os = "visionos",
            target_os = "macos",
            target_os = "tvos",
            target_os = "watchos",
        )
    ))]
```
This means that these methods are only available on macOS-like systems. However, the `IP_BOUND_IF` socket option also exists on SunOS-derived systems (e.g. illumos and Solaris); see for instance [the Solaris `man` page for `ip(7p)`](https://docs.oracle.com/cd/E86824_01/html/E54777/ip-7p.html). We should change the list of systems that these methods are available on so that they also exist on these systems.

hawkw:
The `IP_BOUND_IF` socket option, which is wrapped by the
`Socket::bind_device_by_index_{v4,v6}` and
`Socket::device_index_{v4,v6}` is available on SunOS-like systems, such
as illumos and Solaris, as well as macOS-like systems. However, these
APIs are currently cfg-flagged to only be available on macOS-like
systems.

This commit changes the cfg attributes to also enable these APIs on
illumos and Solaris.

Fixes #560

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -1392,25 +1392,31 @@ pub(crate) fn original_dst_ipv6(fd: Socket) -> io::Result<SockAddr> {
 }
 
 // TODO(eliza): use libc's definition on solarish once it merges...
-#[cfg(any(target_os = "illumos", target_os = "solaris",))]
+#[cfg(all(feature = "all", any(target_os = "illumos", target_os = "solaris")))]
 const IP_BOUND_IF: libc::c_int = 0x41;
-#[cfg(any(
-    target_os = "ios",
-    target_os = "visionos",
-    target_os = "macos",
-    target_os = "tvos",
-    target_os = "watchos",
+#[cfg(all(
+    feature = "all",
+    any(
+        target_os = "ios",
+        target_os = "visionos",
+        target_os = "macos",
+        target_os = "tvos",
+        target_os = "watchos",
+    )
 ))]
 use libc::IP_BOUND_IF;
 
-#[cfg(any(target_os = "illumos", target_os = "solaris",))]
+#[cfg(all(feature = "all", any(target_os = "illumos", target_os = "solaris")))]
 const IPV6_BOUND_IF: libc::c_int = 0x41;
-#[cfg(any(
-    target_os = "ios",
-    target_os = "visionos",
-    target_os = "macos",
-    target_os = "tvos",
-    target_os = "watchos",
+#[cfg(all(
+    feature = "all",
+    any(
+        target_os = "ios",
+        target_os = "visionos",
+        target_os = "macos",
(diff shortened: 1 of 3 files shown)


hawkw:
Ugh, it looks like `libc` also doesn't expose these socket options on illumos or Solaris. That's annoying...

src/sys/unix.rs
```diff
@@ -1389,6 +1391,36 @@ pub(crate) fn original_dst_ipv6(fd: Socket) -> io::Result<SockAddr> {
     .map(|(_, addr)| addr)
 }
 
+// TODO: remove this once https://github.com/rust-lang/libc/pull/4287 merges
+#[cfg(all(feature = "all", any(target_os = "illumos", target_os = "solaris")))]
+const IP_BOUND_IF: libc::c_int = 0x41;
```
hawkw, line 1396: This is a bit of a bummer; I have an upstream PR rust-lang/libc#4287 to add these constants in `libc`, but wanted to test the change without that. I don't have a strong opinion on whether it's important to wait to pick these up from `libc` or if it's okay to just merge with a local definition and use them from libc in the future, but I imagine the maintainers will have an opinion on this.

Thomasdezeeuw: The libc pr is already merged, so let's wait for a release. I really don't want to maintain constants :)

hawkw: Yeah, I'm happy to wait --- this was just intended to allow testing the change temporarily. Once there's a new libc release, I'll pick it up and update this.

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @hawkw

hawkw:
@Thomasdezeeuw would it be possible to get a patch release of `socket2` 0.5.x with this change? I see that the `master` branch has breaking changes for v0.6 (as of 0cb51192f91f63c31373b8bd6cd8aeeddcbc610c), but I'd be happy to fork v0.5.x from the prior commit and do a backport, if that's helpful.

Thomasdezeeuw:
@hawkw yes, if you can create a pr to target 0.5.x that would be helpful 👍

hawkw:
> @hawkw yes, if you can create a pr to target 0.5.x that would be helpful 👍

Opened #565 targeting v0.5.x!

PR #565, opened 2025-03-17
Enable `IP_BOUND_IF` on illumos and Solaris

### LINKED ISSUE #560: `Socket::bind_device_by_index_{v4,v6}` should be available on illumos/Solaris

hawkw:
The `IP_BOUND_IF` socket option, which is wrapped by the `Socket::bind_device_by_index_{v4,v6}` and
`Socket::device_index_{v4,v6}` is available on SunOS-like systems, such as illumos and Solaris, as well as macOS-like systems. However, these APIs are currently cfg-flagged to only be available on macOS-like systems.

This commit changes the cfg attributes to also enable these APIs on illumos and Solaris.

Fixes #560

This is a backport of PR #561 to the `v0.5.x` branch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ targets = ["aarch64-apple-ios", "aarch64-linux-android", "x86_64-apple-darwin", 
 features = ["all"]
 
 [target."cfg(unix)".dependencies]
-libc = "0.2.150"
+libc = "0.2.171"
 
 [target.'cfg(windows)'.dependencies.windows-sys]
 version = "0.52"
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -22,6 +22,8 @@ use std::net::{Ipv4Addr, Ipv6Addr};
         target_os = "macos",
         target_os = "tvos",
         target_os = "watchos",
+        target_os = "illumos",
+        target_os = "solaris",
     )
 ))]
 use std::num::NonZeroU32;
@@ -2076,6 +2078,8 @@ impl crate::Socket {
             target_os = "macos",
             target_os = "tvos",
             target_os = "watchos",
+            target_os = "illumos",
+            target_os = "solaris",
         )
     ))]
     #[cfg_attr(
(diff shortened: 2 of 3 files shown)


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @hawkw

hawkw:
Ah, whoops --- it looks like I accidentally opened this PR against `master` rather than `v0.5.x`, although it was based on `v0.5.x` in my fork. My bad! There are now two copies of this commit on `master` (one of which is a 0-line diff) and none on `v0.5.x`:
image

I've opened a new PR #566 which is actually based on `v0.5.x` the way it should be. You may also want to rewrite history to remove the empty second commit to `master` (33291e25b06d6af82ec05db988a7392fdf24f4da) though I'm not sure how you feel about that sort of thing. Sorry for the inconvenience!

Thomasdezeeuw:
Why does GitHub even allow that? 🤦

hawkw:
I think that GitHub *will* attempt to prevent you from accidentally opening a PR for a commit that was already merged, but since the source branch was based on the `v0.5.x` branch on my fork, it may have understood this PR as a totally separate, new commit, rather than the same commit back for a second time? I'm not sure, though. Oh well.

PR #572, opened 2025-04-22
Added bind_device_by_index_{v4,v6} for linux and android (#569)

boris-n:


src/sys/unix.rs
```diff
@@ -1847,6 +1849,29 @@ impl crate::Socket {
...
+    /// If a socket is bound to an interface, only packets received from that
+    /// particular interface are processed by the socket.
+    ///
+    /// If `interface` is `None`, the binding is removed. If the `interface`
+    /// index is not valid, an error is returned.
+    ///
+    /// One can use [`libc::if_nametoindex`] to convert an interface alias to an
+    /// index.
+    #[cfg(all(feature = "all", any(target_os = "linux", target_os = "android",)))]
+    pub fn bind_device_by_index_v4(&self, interface: Option<NonZeroU32>) -> io::Result<()> {
```
Thomasdezeeuw, line 1863: I don't think we should have two function with the same name that use different socket options.

Either we need to ensure that the socket options are the same, then we can merge the functions, or we need a different function name.

Thomasdezeeuw (changes requested):

src/sys/unix.rs
boris-n: They are the same (for different target_os), I merged into a single functions.

ningyuan-wang:
For Android and Linux, the syscalls are independent of the V4 or V6 stack used, therefore the function name should be `bind_device_by_index`, and not having the suffix attached.

Since `setsockopt` for `SO_BINDTOIFINDEX` is added, the corresponding `getsockopt` is also needed.

Android support required a bump to `libc` version: See [0.2.172](https://github.com/rust-lang/libc/releases/tag/0.2.172) updates for `libc`.

[PR 573](https://github.com/rust-lang/socket2/pull/573) is created that resolves the issues listed.

boris-n:
> For Android and Linux, the syscalls are independent of the V4 or V6 stack used, therefore the function name should be `bind_device_by_index`, and not having the suffix attached.

The question here is mainly if the socket2 library should provide an abstraction from the target OS (as far as possible) or provide a different set of functions for each target OS.

I leave it to the maintainers of the socket2 library to make this decision.

> Since `setsockopt` for `SO_BINDTOIFINDEX` is added, the corresponding `getsockopt` is also needed.

Added. (Again, a decision from the socket2 maintainers is needed if this should be added to the existing `device_index_{v4,v6}` functions, or if a new, IP version independent function should be added for linux/android. 

> Android support required a bump to `libc` version: See [0.2.172](https://github.com/rust-lang/libc/releases/tag/0.2.172) updates for `libc`.

Updated.

Thomasdezeeuw (approved):
Sorry for leaving this so long. I've made a small change to mention `SO_BINDTOIFINDEX` in the getter function, similar to what you already did in the setters.

Thomasdezeeuw:
Thanks @boris-n

PR #580, opened 2025-05-24
Remove libc from public API

Thomasdezeeuw:


src/sys/unix.rs
```diff
@@ -83,7 +83,7 @@ use crate::{Domain, Protocol, SockAddr, SockAddrStorage, TcpKeepalive, Type};
 #[cfg(not(target_os = "redox"))]
 use crate::{MsgHdr, MsgHdrMut, RecvFlags};
 
-pub(crate) use libc::c_int;
+pub(crate) use std::ffi::c_int;
```
Thomasdezeeuw, line 86: We can also keep this as is and add it as an exception.

Thomasdezeeuw:
I'm removing the last bit of libc API on Linux in #581.

Thomasdezeeuw:
This became a bit of a messy pr cleaning up a number of small issues, but at least the commits are clean.

Berrysoft (approved):
LGTM. But I'm not a collaborator:)

Issue #458, opened 2023-07-25
Set Socket interface using interface index


jorritolthuis:
In C, one can set a socket's interface using `setsockopt(socket, SOL_IP, IP_MULTICAST_IF, &mreqn, sizeof(mreqn))` where `mreqn` is
```
struct ip_mreqn {
    struct in_addr imr_multiaddr; /* IP multicast group address */
    struct in_addr imr_address;   /* IP address of local interface */
    int            imr_ifindex;   /* interface index */
};
```

sockets2 offers the `set_multicast_if_v4` function, but only accepts an address as parameter. I would like to set the interface using sockets2 based on an interface index.

Thomasdezeeuw:
A pr for this would be welcome. We'll have to look at how the `ip_mreqn` type would look as it needs to be cross platform(-ish). Another point is the migration of the API as we can't have breaking changing, but in future version we'll want a single function for `IP_MULTICAST_IF`

PR #582, opened 2025-05-24
Added Socket::set_multicast_if_v4_n to set interface by index (#458)

boris-n:
Implementation for Socket::set_multicast_if_v4_n (similar to Socket::join_multicast_v4_n) which allows to set interface by index or address (via struct InterfaceIndexOrAddress), for Linux/Android/Fuchsia/FreeBSD/NetBSD. (Other OSes seem to have no support for this.)

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -845,7 +845,6 @@ fn set_common_accept_flags(socket: Socket) -> io::Result<Socket> {
 #[cfg(not(any(
     target_os = "haiku",
     target_os = "illumos",
-    target_os = "netbsd",
     target_os = "redox",
     target_os = "solaris",
 )))]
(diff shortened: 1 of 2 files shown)


boris-n:
Added feature = "all" as setting multicast interface by index is not supported on Windows and MacOSX.

michaelinoue:
Keen for this to merge!

michaelinoue:
Any thoughts on expanding this to set both an address and an if index? It seems like we're pinned to using either or in this PR, but we could potentially want both (e.g. sending multicast traffic from a particular address on the interface).

src/socket.rs
```diff
@@ -1480,6 +1479,54 @@ impl Socket {
...
+            // byte order)
+            let addr = match interface {
+                InterfaceIndexOrAddress::Index(index) => {
+                    if *index >= 0x0100_0000 {
+                        return Err(io::Error::new(
+                            io::ErrorKind::AddrNotAvailable,
+                            "Interface index out of bounds",
+                        ));
+                    }
+                    Ipv4Addr::from_bits(*index)
```
Thomasdezeeuw: This doesn't seem right. The enum variant clearly say `Index`, but we're using an IPv4 address. I think we shouldn't support the index here and recommend to use the `Address` instead for NetBSD, but I'm not sure.

Thomasdezeeuw (commented):
I left a comment, but I'm not sure about this pr. I don't use these APIs and currently don't have the time to look into them properly. For this to be merged it would be nice if someone with experience with the options could confirm they work as expected. This also needs a test.

src/socket.rs
boris-n: The implementation is correct (although I have to admit it is probably confusing). The NetBSD API seems to be a bit "unusual" for setting the multicast interface by index. Let me cite from the NetBSD man page (https://man.netbsd.org/NetBSD-7.0/ip.4):

> The IP_MULTICAST_IF option overrides the default for subsequent transmissions from a given socket:
>    struct in_addr addr;
>    setsockopt(s, IPPROTO_IP, IP_MULTICAST_IF, &addr, sizeof(addr));
> where "addr" is the local IP address of the desired interface [...]. An application may also specify an alternative to the default network interface by index:
>    struct uint32_t idx = htonl(ifindex);
>    setsockopt(s, IPPROTO_IP, IP_MULTICAST_IF, &idx, sizeof(idx));
> where "ifindex" is an interface index as returned by if_nametoindex(3).

So, while Linux, FreeBSD, Android and Fuchsia use a special `struct mreqn` (which has a separate field for the interface index) if they want to set the multicast interface by index, NetBSD uses a `struct in_addr` to set the address by IP and an `uint32_t` (in network byte order.) to set the address by interface index. `struct in_addr` and `uint32_t` both are 4 bytes in size, so the kernel cannot distinguish the two cases from the size parameter. Looking into the kernel code shows that IP addresses in subnet 0.0.0.0/8 are interpreted as interface index (see https://github.com/NetBSD/src/blob/trunk/sys/netinet/ip_output.c#L1691). This is the reason why I converted the interface to an address and used this one with the `setsockopt` call.

But I agree, the implementation is probably easier to understand if the code aligns to the examples given in the NetBSD manual page. I provided a modified version which does so.

boris-n: But NetBSD has support for setting the multicast interface by index, they documented in the man page that you have to pass it as `uint32_t` in network byte order (as my code now does).
There is no standardized way for setting the multicast interface by index, and some OSes don't support it. But Linux supports it using `struct mreq` for `IP_MULTICAST_IF`, and NetBSD supports it using `uint32_t` in network byte order for `IP_MULTICAST_IF` - that's totally fine. So why not support it for NetBSD? Only because NetBSD doesn't do it the Linux way using `struct ip_mreqn`?
(I agree that the NetBSD way is sub-optimal, the API will not be able to bind to an address in 0.0.0.0/8 address. But that's more an issue with the binding to an address instead of binding to an interface index. And as long as this subnet stays reserved, there is no issue with the NetBSD API.)

boris-n: You need to read the man page (https://man.netbsd.org/ip.4) further down where the man page specifies how to set the network interface by index instead of by address:

    An application may also specify an alternative to the default network interface by index:

     struct uint32_t idx = htonl(ifindex);
     setsockopt(s, IPPROTO_IP, IP_MULTICAST_IF, &idx, sizeof(idx));

     where "ifindex" is an interface index as returned by [if_nametoindex(3)](https://man.netbsd.org/if_nametoindex.3).

This matches what is in the branch in Rust for NetBSD:

    let index_be = (*index as u32).to_be();
    unsafe {
        setsockopt(
            self.as_raw(),
            sys::IPPROTO_IP,
            sys::IP_MULTICAST_IF,
            index_be,
        )
    }

Issue #575, opened 2025-04-30
Add support for sendmmsg


Hasan6979:
Right now socket2 only supports sendmsg, it would be nice if it also supports sendmmsg

Issue #574, opened 2025-04-30
Add support for recvmmsg


Hasan6979:
Current socket2 crate only supports recvmsg, it would be great if multiple messages were able to send through with a single syscall.

PR #583, opened 2025-05-24
Added sendmmsg and recvmmsg (#574) (#575)

boris-n:
Added sendmmsg and recvmmsg methods for Android, Linux, FreeBSD, NetBSD, OpenBSD and AIX. Note that Windows has no support for this.
Note that the timeout parameter in recvmmsg is buggy on Linux (see BUGS section in the recvmmsg manual page on Linux), but it should work on *BSD.
This also adds MMsgHdr and MMsgHdrMut structs (similar to MsgHdr and MsgHdrMut which are used for sendmsg/recvmsg). It is a bit nasty that I had to copy the methods of MsgHdr/MsgHdrMut to MMsgHdr/MMsgHdrMut but I don't see a better way.
I also added a simple unittest for sendmmsg and recvmmsg.


boris-n:
Added feature = "all" as sendmmsg/recvmmsg are not available on Windows and MacOSX.

Issue #587, opened 2025-05-26
Add `set_priority` for SO_PRIORITY


kleinesfilmroellchen:
It would be useful if the socket option SO_PRIORITY (like SO_LINGER and many others) was available directly, without having to resort to libc functions.

PR #588, opened 2025-06-01
Added SO_PRIORITY socket option (#587)

boris-n:


src/socket.rs
```diff
@@ -996,6 +996,35 @@ impl Socket {
         }
     }
 
+    /// Get value for the `SO_PRIORITY` option on this socket.
+    ///
+    /// For more information about this option, see [`set_priority`].
+    ///
+    /// [`set_priority`]: Socket::set_priority
+    #[cfg(any(target_os = "linux", target_os = "android", target_os = "fuchsia"))]
```
Thomasdezeeuw, line 1004: This needs the `all` feature if it's not available on Windows and macOS as well. See the `cfg` that uses `features = "all"` on some other methods.

Thomasdezeeuw (changes requested):
Can we have a test for this?

Thomasdezeeuw (approved):

Issue #564, opened 2025-03-14
only_v6 panic on windows


BiagioFesta:
This sample code
```
use socket2::SockRef;
use std::net::{Ipv4Addr, SocketAddr, UdpSocket};

fn is_only_v6(socket: SockRef) -> Option<bool> {
    socket.only_v6().ok()
}

fn main() {
    let socket = UdpSocket::bind(SocketAddr::new(Ipv4Addr::LOCALHOST.into(), 0)).unwrap();
    println!("{:?}", is_only_v6(SockRef::from(&socket)));
}
```

panics (debug assertion) on Windows
```
thread 'main' panicked at ..\socket2-0.5.8\src\sys\windows.rs:772:9:
assertion `left == right` failed
  left: 1
 right: 4
```

Is this expected and I am missing something?

Thomasdezeeuw:
I think Windows doesn't use a `c_int` for this option, but a boolean (the 1 byte one, not the 4 byte on they sometimes use). So we need to update the type: https://github.com/rust-lang/socket2/blob/5c1f3b09897d251b8a10f690af71af13cc7d32ca/src/socket.rs#L1880.

Weird that our testing never caught this before.

Thomasdezeeuw:
Hey @BiagioFesta I tried changing the type to 1 byte (from 4), but it fails, see #593, https://github.com/rust-lang/socket2/actions/runs/16035057825/job/45244713907?pr=593.

Thomasdezeeuw:
Per #593, the 4 byte boolean is not correct, so closing this.

BiagioFesta:
> In https://github.com/rust-lang/socket2/pull/593 we determined that using 4 bytes is incorrect.

I understand this, but beside the possible root cause (which might have been rightly or wrongly identified), and a possible fix the issue remains: `socket.only_v6().ok()` crashes (quite consistently) on windows.

Is the assertion firing an expectation of the library behavior? 

So I am still wondering if we need to re-open this issue (and why has been closed)

ChrisDenton:
I personally do agree the assertion is too restrictive. As I said on the open PR, it's perfectly legal (according to the API docs) for the OS to write less than the full size given. The returned value in this case must be an integer whose size is <= 4 bytes (so 1, 2 or 4) but beyond that there aren't any guarantees.

BiagioFesta:
Hi, I'd like to respectfully revisit this since my core concern wasn't addressed before closing.

[...]

PR #603, opened 2025-07-06
fix only_v6() panic on windows

zh-jq:
I have encountered a panic case in only_v6 in github CI on windows:
https://github.com/bytedance/g3/actions/runs/16095794585/job/45418469197

The input buffer size (optlen) must be 4, and the output optlen may be 1. As the input optval is initialized to zero, it will be modified to nonzero when ipv6 only, so just use `optval != 0` to get the result value.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -1932,10 +1932,13 @@ impl Socket {
     ///
     /// [`set_only_v6`]: Socket::set_only_v6
     pub fn only_v6(&self) -> io::Result<bool> {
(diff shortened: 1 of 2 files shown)


zh-jq:
> We tried this already, see #593 and #564, it fails on our Windows CI.

It seems that the input buffer size must be 4 bytes, and the output optlen may be 1, so the check could be: it will be ipv6 only if all the buffer bits are zero.

Thomasdezeeuw:
Fixed by #655.

PR #622, opened 2025-11-16
fix getsockopt generic argument in notsent_lowat

the8472:
small cleanup after #611
I made a copy-paste error there: notsent_lowat is not a boolean but a value. `Bool` is an alias for `c_int` so this doesn't make a real difference, but it's semantically the wrong type and might confuse future readers.


the8472:
CI failure will be fixed by https://github.com/rust-lang/rust/pull/148914

Thomasdezeeuw:
Thanks @the8472

PR #627, opened 2025-12-19
CI: test MSRV on windows, too

hanna-kruppe:
Noticed in #621 that this isn't tested

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,13 +14,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        build: [stable, 1.70.0, beta, nightly, macos, windows]
+        build: [stable, msrv-linux, msrv-windows, beta, nightly, macos, windows]
         include:
           - build: stable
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -14,14 +14,17 @@ jobs:
...
       fail-fast: false
       matrix:
-        build: [stable, 1.70.0, beta, nightly, macos, windows]
+        build: [stable, msrv-linux, msrv-windows, beta, nightly, macos, windows]
         include:
           - build: stable
             os: ubuntu-latest
             rust: stable
-          - build: 1.70.0
+          - build: msrv-linux
```
hanna-kruppe: NB: I had to come up with a new value for `build` because using the same for both MSRV jobs means only one of them runs. The github UI on other PRs suggests that this job wasn't a "required check" even when it was called "Test (1.70.0)", but in case the MSRV check is intended to be required, someone will have to dig through the branch protection rules to make them required under the new names.

PR #630, opened 2025-12-28
Fix multicast_ttl_v4 on non-Linux

jtojnar:
On Windows, `set_multicast_ttl_v4` fails with:

```
Os { code: 10022, kind: InvalidInput, message: "An invalid argument was supplied." }
```

On MacOS and FreeBSD, `multicast_ttl_v4` fails with:

```
Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
```

According to [FreeBSD’s ip(4)](https://man.freebsd.org/cgi/man.cgi?query=ip&apropos=0&sektion=0&manpath=FreeBSD+15.0-RELEASE+and+Ports&arch=default&format=html), it is supposed to be u8:

> ```c
> u_char ttl;     /* range: 0 to 255, default = 1 */
> setsockopt(s, IPPROTO_IP, IP_MULTICAST_TTL, &ttl, sizeof(ttl));
> ```

I do not understand why it fail on [Windows](https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options), where it should be `DWORD` (i.e. `u32`).

### DIFF
--- a/tests/socket.rs
+++ b/tests/socket.rs
@@ -1352,36 +1352,40 @@ macro_rules! test {
         #[test]
         $( #[$attr] )*
         fn $get_fn() {
-            test!(__ Domain::IPV4, STREAM, $get_fn, $set_fn($arg), $expected);
+            test!(__ Domain::IPV4, STREAM, None, $get_fn, $set_fn($arg), $expected);
             #[cfg(not(target_os = "vita"))]
-            test!(__ Domain::IPV6, STREAM, $get_fn, $set_fn($arg), $expected);
+            test!(__ Domain::IPV6, STREAM, None, $get_fn, $set_fn($arg), $expected);
         }
     };
     // Only test using a IPv4 socket.
     (IPv4:DGRAM $get_fn: ident, $set_fn: ident ( $arg: expr ) ) => {
         #[test]
         fn $get_fn() {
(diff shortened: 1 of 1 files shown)


jtojnar:
Actually, `set_multicast_ttl_v4` [fails](https://github.com/rust-lang/socket2/actions/runs/20551407174/job/59029589312?pr=630) on Linux as well:

    Os { code: 22, kind: InvalidInput, message: "Invalid argument" }

Even though it succeeds in [tokio](https://github.com/tokio-rs/tokio/pull/7781), which uses essentially the same code internally:
https://github.com/rust-lang/rust/blob/4f834264ab8d0cb95ac7972532661c84a154667e/library/std/src/sys/net/connection/socket.rs#L639

jtojnar:
Right.

I had a theory that the tests were failing on Darwin because of wrong size of the value passed to `setsockopt` but it looks like it actually passes even with `u32` hardcoded (reverted the fix in second commit to verify). I just used wrong socket type (TCP instead of UDP) in the test initially.

And it also turned out that tokio’s `UdpSocket` does not even use `socket2` for this option – it uses `std`, which properly uses `u8` on non-Linux UNIXes. And Windows was correct all along in both stacks.

So I am even more confused why the tests in tokio were failing on Windows and non-Linux UNIXes with the same errors like here initially:
https://github.com/tokio-rs/tokio/pull/7781#discussion_r2649568710

PR #632, opened 2026-01-02
Don't set SO_NOSIGPIPE when accepting sockets

### LINKED ISSUE #631: Using accept on macOS unix socket always raise EINVAL
`accept_raw` is fine.

This happens because `SO_NOSIGPIPE` is set for `accept`.

If this is not a bug, I'd like to know if this is a design decision. Then adding a note on macOS line "use accept_raw for unix socket" will be helpful.

```rust
use socket2::{Domain, Socket, Type, SockAddr};
use std::thread;

fn main() {
    let path = "/tmp/test.sock";
    let _ = std::fs::remove_file(path);

    let server = Socket::new(Domain::UNIX, Type::STREAM, None).unwrap();
    server.bind(&SockAddr::unix(path).unwrap()).unwrap();
    server.listen(1).unwrap();

    thread::spawn(move || {
        Socket::new(Domain::UNIX, Type::STREAM, None)
            .unwrap()
            .connect(&SockAddr::unix(path).unwrap())
            .unwrap();
    });

    server.accept().expect("macOS always fail here");
}
```

Thomasdezeeuw:
This is inherited from the parent socket. Furthermore when attempting to set it on a Unix socket domain it will result in an error.

Closes #631

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -762,6 +762,8 @@ const fn set_common_type(ty: Type) -> Type {
 }
 
 /// Set `FD_CLOEXEC` and `NOSIGPIPE` on the `socket` for platforms that need it.
+///
+/// Sockets created via `accept` should use `set_common_accept_flags` instead.
 fn set_common_flags(socket: Socket) -> io::Result<Socket> {
     // On platforms that don't have `SOCK_CLOEXEC` use `FD_CLOEXEC`.
     #[cfg(all(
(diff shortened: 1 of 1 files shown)


src/socket.rs
```diff
@@ -798,6 +796,34 @@ fn set_common_flags(socket: Socket) -> io::Result<Socket> {
     Ok(socket)
 }
 
+/// Set `FD_CLOEXEC` on the `socket` for platforms that need it.
+///
+/// Unlike `set_common_flags` we don't set `NOSIGPIPE` as that is inherited.
+/// Furthermore attempts to set it on a unix socket domain results in an error.
```
Darksonn:
```suggestion
/// Unlike `set_common_flags` we don't set `NOSIGPIPE` as that is inherited from the listener.
/// Furthermore, attempts to set it on a unix socket domain results in an error.
```

PR #635, opened 2026-01-13
make `MsgHdr` and `MsgHdrMut` `repr(transparent)`

### LINKED ISSUE #633: `MsgHdr[Mut]` should be `repr(transparent)`
Since documentation explicitly states
> This wraps msghdr on Unix and WSAMSG on Windows

`MsgHdr[Mut]` should be repr(transparent) over this type, or, at least, provide:
- `impl From<MsgHdr[Mut]> for libc::msghdr`
- `MsgHdr[Mut]::as_hdr_ref() -> &libc::msghdr`
- `MsgHdrMut::as_hdr_mut() -> &mut libc::msghdr`

I can start working on this if it's OK

nerodono:
Closes #633

This PR simply adds `#[repr(transparent)]` to corresponding types, without any methods for conversion on the types themselves.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -562,6 +562,7 @@ impl TcpKeepalive {
 /// This wraps `msghdr` on Unix and `WSAMSG` on Windows. Also see [`MsgHdrMut`]
 /// for the variant used by `recvmsg(2)`.
 #[cfg(not(target_os = "redox"))]
+#[repr(transparent)]
 pub struct MsgHdr<'addr, 'bufs, 'control> {
     inner: sys::msghdr,
     #[allow(clippy::type_complexity)]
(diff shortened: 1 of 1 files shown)


nerodono:
Wrote improvised changelog, hope it's done the right way. By the way, after the merge, can you please release a patch version on crates.io? On my workplace, we decided to use this library for some low-level socket operations, and proposed fix would be required (practically speaking, it's not, since transmuting `MsgHdr[Mut]` into `libc::msghdr`  won't do anything bad even without `repr(transparent)`, but lack of this guarantee in the API feels wrong) and making git dependency would require to monitor socket2 releases to get rid of it.

Darksonn:
This is ok with me, but I think we usually do not add changelog entries for unreleased changes.

nerodono:
> This is ok with me, but I think we usually do not add changelog entries for unreleased changes.

So, it's OK to simply not add changelog in this PR?

Thomasdezeeuw (approved):
You can remove the part for the changelog, otherwise LGTM.

PR #639, opened 2026-02-11
Add wasm32-wasip2 target support

### LINKED ISSUE #268: Support wasm (in some form)

ricochet:
Rework of #532 by @nickrum, addressing review feedback from @Darksonn
and @Thomasdezeeuw:

- Reuses sys/unix.rs instead of creating sys/wasi.rs (per @Thomasdezeeuw's
  code review feedback)
- MaybeUninitSlice stays unconditional since libc::iovec is available on WASI,
  eliminating 11+ scattered cfg gates (per @Darksonn's concern)
- Internal cfg changes concentrated in the sys module

Opt-in cfg patterns use `all(target_os = "wasi", not(target_env = "p1"))`
to match how the libc crate gates WASI p2 socket support. Opt-out
patterns use bare `target_os = "wasi"` for features unavailable on any
WASI version.

WASI tests run in CI with wasmtime (27/27 pass). Tests for unsupported
socket options are gated out with references to the relevant WASI spec
issues (SO_BROADCAST, SO_LINGER, IPV6_V6ONLY).

```bash
export CARGO_TARGET_WASM32_WASIP2_RUNNER="wasmtime --wasi inherit-network"
# Doc-tests are excluded because the SockRef example spawns threads, which
# WASI does not yet support.
cargo test --target wasm32-wasip2 --lib --tests
```

Closes #268

Based-on: https://github.com/rust-lang/socket2/pull/532
Co-authored-by: Nicola Krumschmidt <git@nkcom.de>

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -289,8 +289,8 @@ pub(crate) use libc::{IPV6_MULTICAST_HOPS, IPV6_MULTICAST_IF, IP_MULTICAST_IF, M
         target_os = "netbsd",
         target_os = "tvos",
         target_os = "watchos",
-        all(target_os = "wasi", not(target_env = "p1")),
         target_os = "cygwin",
+        all(target_os = "wasi", not(target_env = "p1")),
     )
 ))]
 pub(crate) use libc::{TCP_KEEPCNT, TCP_KEEPINTVL};
@@ -450,8 +450,8 @@ impl Type {
             target_os = "linux",
             target_os = "netbsd",
             target_os = "openbsd",
-            all(target_os = "wasi", not(target_env = "p1")),
             target_os = "cygwin",
+            all(target_os = "wasi", not(target_env = "p1")),
         )
     ))]
     pub const fn nonblocking(self) -> Type {
@@ -1261,8 +1261,8 @@ pub(crate) fn set_tcp_keepalive(fd: RawSocket, keepalive: &TcpKeepalive) -> io::
         target_os = "netbsd",
         target_os = "tvos",
         target_os = "watchos",
-        all(target_os = "wasi", not(target_env = "p1")),
         target_os = "cygwin",
+        all(target_os = "wasi", not(target_env = "p1")),
     ))]
     {
         if let Some(interval) = keepalive.interval {
--- a/tests/socket.rs
+++ b/tests/socket.rs
@@ -542,7 +542,9 @@ fn connect_timeout_unrouteable() {
 }
 
 #[test]
-#[cfg(not(any(target_os = "vita", target_os = "wasi")))] // WASI uses start-connect/finish-connect, no connect_timeout.
+// Loopback has special behavior on vita.
+// WASI uses start-connect/finish-connect, no connect_timeout.
+#[cfg(not(any(target_os = "vita", target_os = "wasi")))]
 fn connect_timeout_unbound() {
     // Bind and drop a socket to track down a "probably unassigned" port.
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -49,7 +49,7 @@ targets = [
 [package.metadata.playground]
 features = ["all"]
 
-[target."cfg(unix)".dependencies]
+[target.'cfg(any(unix, target_os = "wasi"))'.dependencies]
 libc = "0.2.172"
```
Darksonn: What libc versions is required as a minimum?

src/sys/unix.rs
```diff
@@ -425,6 +450,7 @@ impl Type {
             target_os = "linux",
             target_os = "netbsd",
             target_os = "openbsd",
+            all(target_os = "wasi", not(target_env = "p1")),
             target_os = "cygwin",
```
Thomasdezeeuw: Nit.
```suggestion
            target_os = "cygwin",
            all(target_os = "wasi", not(target_env = "p1")),
```

src/sockaddr.rs
```diff
@@ -350,7 +355,10 @@ impl From<SocketAddrV4> for SockAddr {
             storage.sin_family = AF_INET as sa_family_t;
             storage.sin_port = addr.port().to_be();
             storage.sin_addr = crate::sys::to_in_addr(addr.ip());
-            storage.sin_zero = Default::default();
+            #[cfg(not(target_os = "wasi"))]
+            {
+                storage.sin_zero = Default::default();
+            }
```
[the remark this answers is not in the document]
Darksonn: Pretty sure even latest stable does not let you get rid of this block.

Cargo.toml
ricochet: The minimum libc version for the WASI socket2-specific types is 0.2.162 (added in [libc PR #3981](https://github.com/rust-lang/libc/pull/3981), which was created specifically for socket2's wasip2 support). The version wasn't changed by this PR — it was already 0.2.172 on master.

tests/socket.rs
```diff
@@ -530,7 +542,7 @@ fn connect_timeout_unrouteable() {
 }
 
 #[test]
-#[cfg(not(target_os = "vita"))] // Loopback has special behavior on vita
```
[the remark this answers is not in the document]
ricochet: Good catch — restored it. The comment now reads: `// Loopback has special behavior on vita. WASI uses start-connect/finish-connect, no connect_timeout.`

src/sockaddr.rs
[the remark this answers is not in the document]
ricochet: Right, `#[cfg]` on expressions is not stabilized (as Darksonn confirmed), so the block is needed regardless of MSRV.

src/sys/unix.rs
ricochet: Fixed — swapped the ordering in all three locations where `all(target_os = "wasi", not(target_env = "p1"))` appeared before `target_os = "cygwin"` (lines 292, 453, and 1264).

Thomasdezeeuw:
Thanks @ricochet

I'm merging this now. @Darksonn if you want any changes regarding https://github.com/rust-lang/socket2/pull/639#discussion_r2802952868 let me know.

Darksonn:
No it sounds ok.

PR #645, opened 2026-03-20
feat: add ancillary data (control message) helpers for sendmsg/recvmsg

### LINKED ISSUE #614: Support for ancillary data/control messages in `sendmsg`/`recvmsg` as well as the upcoming `sendmmsg`/`recvmmsg`
The idiomatic way of parsing and constructing ancillary data in C is to use the various macros defined in `sys/socket.h` (`CMSG_FIRSTHDR`, `CMSG_NXTHDR` etc.). Those macros act on `msghdr` directly. Since we no longer return `msghdr` in the API, we should serialize/deserialize these messages for the user in the API as well.

seuros:
Closes rust-lang/socket2#614

Add Unix-only safe wrappers for CMSG_* operations:

- `cmsg_space(data_len) -> Option<usize>`: compute control buffer size (returns None when data_len overflows c_uint, fixing a truncation path that would produce an undersized buffer)
- `ControlMessage<'a>`: a parsed ancillary data entry (level, type, data)
- `ControlMessages<'a>`: iterator over a received control buffer; walks via byte-offset arithmetic and ptr::read_unaligned so no aligned reference to cmsghdr is ever formed (avoids UB on 1-byte-aligned Vec<u8> buffers)
- `ControlMessageEncoder<'a>`: builder for outgoing control messages; rejects payloads exceeding c_uint::MAX before calling CMSG_SPACE/CMSG_LEN to prevent buffer overflow from silent truncation

Also add a Cross CI job to run tests under QEMU on i686, aarch64, and armv7 Linux,  the target families where CMSG_ALIGN factor and cmsg_len width differ from x86_64.

Enables SCM_RIGHTS file-descriptor passing without depending on libc directly (tracked by plabayo/rama#781).

I can remove the Cross tests if needed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,7 +38,7 @@ jobs:
             os: windows-latest
             rust: stable
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - uses: dtolnay/rust-toolchain@master
       with:
         toolchain: ${{ matrix.rust }}
@@ -52,7 +52,7 @@ jobs:
     name: Rustfmt
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - uses: dtolnay/rust-toolchain@stable
       with:
         components: rustfmt
@@ -110,7 +110,7 @@ jobs:
           - x86_64-unknown-redox
           - wasm32-wasip2
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - uses: dtolnay/rust-toolchain@nightly
       with:
         components: rust-src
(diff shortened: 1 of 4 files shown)


seuros:
The only remaining CI failure is **FreeBSD (Cirrus CI)** — this is an infrastructure flake, not a code issue:

```
error: failed to get `libc` as a dependency of package `socket2 v0.6.3`
Caused by: download of config.json failed
Caused by: failed to download from `https://index.crates.io/config.json`
Caused by: [60] SSL peer certificate or SSH remote key was not OK
  (SSL certificate OpenSSL verify result: unable to get local issuer certificate (20))
```

Cargo tries to update the crates.io index and hits an SSL cert verification failure in the Cirrus FreeBSD 14.3 VM. Unrelated to this PR! Verified manually on FreeBSD 16-Current where `cargo build`, `cargo build --no-default-features`, and all cmsg tests pass cleanly.

.github/workflows/main.yml
Thomasdezeeuw, line 1: Can you drop the CI changes from this pr? Let's focus this pr on one thing.

.github/workflows/main.yml
```diff
@@ -119,11 +119,44 @@ jobs:
       run: cargo hack check -Z build-std=std,panic_abort --feature-powerset --target ${{ matrix.target }}
     - name: Check docs
       run: RUSTDOCFLAGS="-D warnings --cfg docsrs" cargo doc -Z build-std=std,panic_abort --no-deps --all-features --target ${{ matrix.target }}
+  Cross:
+    name: Cross-test (${{ matrix.target }})
```
Thomasdezeeuw, line 123: Why do we need this? Isn't `cargo check` sufficient?

tests/socket.rs
```diff
@@ -1970,3 +1970,183 @@ fn set_busy_poll() {
...
+        pid: unsafe { libc::getpid() },
+        uid: unsafe { libc::getuid() },
+        gid: unsafe { libc::getgid() },
+    };
+    let cred_bytes = unsafe {
+        std::slice::from_raw_parts(
+            &cred as *const libc::ucred as *const u8,
+            std::mem::size_of::<libc::ucred>(),
+        )
+    };
```
Thomasdezeeuw, line 2112: We still need a bunch of unsafe and libc types... wouldn't it be better to have a safe wrapper in the library?

Thomasdezeeuw (changes requested):
Can you open an issue to discuss this feature? I don't think the code as-is is a good fit for socket2. Based on the tests we still need to use unsafe & libc types to actually use this.

.github/workflows/main.yml
seuros: `cargo check` verify only the current host architecture

.github/workflows/main.yml
seuros: yes

tests/socket.rs
seuros: Let me try this.

seuros (commented):
forgot to submit the comments. just checked my prs.

seuros:
@Thomasdezeeuw the issue is already in #614 there are more related issues to it.

Thomasdezeeuw:
> @Thomasdezeeuw the issue is already in #614 there are more related issues to it.

That issue has no discussion on the design. I didn't mean to open an issue just for the sake of opening an issue, I meant that we should discuss the design of larger features like this.

Issue #646, opened 2026-04-04
IPv4 multicast TTL and loopback use a different ABI than std and system headers on BSD-family targets


lopopolo:
## Report provenance

This bug was found by OpenAI Codex and validated by @lopopolo.

## Summary

`socket2` uses `c_int` for `IP_MULTICAST_TTL` and `IP_MULTICAST_LOOP` everywhere.

The BSD system headers mandate unsigned char for these options and this is reflected by Rust std carrying a target-specific ABI split for these exact options to use `c_uchar` on:

- `dragonfly`
- `freebsd`
- `openbsd`
- `netbsd`
- `solaris`
- `illumos`
- `nto`

## Root cause

The IPv4 multicast getters and setters in `socket2` use `c_int` directly for:

- `Socket::multicast_ttl_v4`
- `Socket::set_multicast_ttl_v4`
- `Socket::multicast_loop_v4`
- `Socket::set_multicast_loop_v4`

Rust std instead carries an internal `IpV4MultiCastType` alias and makes it
`c_uchar` on the targets above:

- std source: <https://doc.rust-lang.org/nightly/src/std/sys/net/connection/socket/mod.rs.html>

The platform headers and manuals also point in that direction. For example:

- FreeBSD `in.h` annotates `IP_MULTICAST_TTL` and `IP_MULTICAST_LOOP` as `u_char`
- NetBSD `ip(4)` documents `IP_MULTICAST_TTL` as `u_char ttl` and `IP_MULTICAST_LOOP` as `u_char loop`: <https://man.netbsd.org/NetBSD-9.2/emips/ip.4>

## Validation

I checked this on `FreeBSD 15.0-RELEASE-p5 amd64`. A raw `getsockopt` probe
returned a four-byte payload for both options:

```text
IP_MULTICAST_TTL rc=0 len=4 bytes=01 00 00 00
IP_MULTICAST_LOOP rc=0 len=4 bytes=01 00 00 00
```

So I am not claiming a confirmed security issue on that host.

The bug here is the ABI divergence from std:

- `socket2` and std disagree on the value type for these options
- BSD-family headers/docs suggest a narrower type
- at least one current FreeBSD kernel still returns a 4-byte payload in practice

## Suggested direction

Match std's target-specific alias for these options:

- introduce an internal `IpV4MultiCastType`
- use `u8` / `c_uchar` on `dragonfly`, `freebsd`, `openbsd`, `netbsd`, `solaris`, `illumos`, and `nto`
[...]

PR #647, opened 2026-04-05
Match BSD-family IPv4 multicast TTL/loopback ABI to system headers

lopopolo:
Refs #646.

BSD-family system headers declare `IP_MULTICAST_TTL` and `IP_MULTICAST_LOOP` as `u_char` / `unsigned char`.

`socket2` was using `c_int` for these options on all targets. On FreeBSD 15 this is observable: `std::net::UdpSocket::set_multicast_ttl_v4(258)` succeeds and stores `2`, while `socket2::Socket::set_multicast_ttl_v4(258)` returns `EINVAL`.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -2479,23 +2479,3 @@ from!(net::UdpSocket, Socket);
 from!(Socket, net::TcpStream);
 from!(Socket, net::TcpListener);
 from!(Socket, net::UdpSocket);
-
-#[cfg(test)]
-mod tests {
-    #[test]
-    #[cfg(target_os = "freebsd")]
-    fn set_multicast_ttl_v4_matches_std_for_values_above_u8_range() {
-        use super::Socket;
-        use crate::{Domain, Protocol, Type};
-        use std::net::{Ipv4Addr, UdpSocket};
-
-        let std_socket = UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)).unwrap();
-        std_socket.set_multicast_ttl_v4(258).unwrap();
-        let std_ttl = std_socket.multicast_ttl_v4().unwrap();
-        assert_eq!(std_ttl, 2);
-
-        let socket2_socket = Socket::new(Domain::IPV4, Type::DGRAM, Some(Protocol::UDP)).unwrap();
-        socket2_socket.set_multicast_ttl_v4(258).unwrap();
-        assert_eq!(socket2_socket.multicast_ttl_v4().unwrap(), std_ttl);
-    }
-}
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -16,6 +16,16 @@ use std::net::Ipv6Addr;
 use std::net::{self, Ipv4Addr, Shutdown};
 #[cfg(any(unix, all(target_os = "wasi", not(target_env = "p1"))))]
 use std::os::fd::{FromRawFd, IntoRawFd};
+#[cfg(any(
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "openbsd",
+    target_os = "netbsd",
+    target_os = "solaris",
+    target_os = "illumos",
+    target_os = "nto",
+))]
+use std::os::raw::c_uchar;
 #[cfg(windows)]
 use std::os::windows::io::{FromRawSocket, IntoRawSocket};
 use std::time::Duration;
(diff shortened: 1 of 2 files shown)


src/socket.rs
```diff
@@ -2445,3 +2479,23 @@ from!(net::UdpSocket, Socket);
 from!(Socket, net::TcpStream);
 from!(Socket, net::TcpListener);
 from!(Socket, net::UdpSocket);
+
+#[cfg(test)]
+mod tests {
```
Thomasdezeeuw: Can you move this to the other tests in the tests directory?

src/socket.rs
```diff
@@ -2445,3 +2479,23 @@ from!(net::UdpSocket, Socket);
...
+    #[test]
+    #[cfg(target_os = "freebsd")]
```
Thomasdezeeuw: Why run it on FreeBSD only?

src/socket.rs
```diff
@@ -2445,3 +2479,23 @@ from!(net::UdpSocket, Socket);
...
+    fn set_multicast_ttl_v4_matches_std_for_values_above_u8_range() {
```
Thomasdezeeuw: Can we please fix these terrible AI generated names?

src/socket.rs
```diff
@@ -2445,3 +2479,23 @@ from!(net::UdpSocket, Socket);
...
+        use std::net::{Ipv4Addr, UdpSocket};
+
+        let std_socket = UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)).unwrap();
+        std_socket.set_multicast_ttl_v4(258).unwrap();
+        let std_ttl = std_socket.multicast_ttl_v4().unwrap();
+        assert_eq!(std_ttl, 2);
+
+        let socket2_socket = Socket::new(Domain::IPV4, Type::DGRAM, Some(Protocol::UDP)).unwrap();
+        socket2_socket.set_multicast_ttl_v4(258).unwrap();
+        assert_eq!(socket2_socket.multicast_ttl_v4().unwrap(), std_ttl);
```
Thomasdezeeuw: Can you use the test macro for this? See the other tests.

src/socket.rs
lopopolo: I've renamed the test to `multicast_v4_bsd_abi`.

src/socket.rs
lopopolo: I mentioned this in the comment I left on the original issue https://github.com/rust-lang/socket2/issues/646#issuecomment-4189157980, but in case you missed it:

> The regression test only is gated on for freebsd because that is the VM I happened to have available for reproduction. I can widen this cfg if you'd like so it covers all of the relevant BSD OSes.

I've widened the cfg so it tests on all relevant BSD OSes.

src/socket.rs
lopopolo: thanks, I was wondering why there were so few unit tests in this crate 😅

I've moved this into `tests/socket.rs`.

src/socket.rs
lopopolo: I've factored this out into an assertion helper named `assert_multicast_ttl_v4`.

lopopolo: To maintain the part of the test that ensures behavior matches std, the helper accepts either a socket2 `Socket` or a std `UdpSocket` via a `SocketRef` enum which I am using to model an "either" enum. If this isn't acceptable, please let me know what you'd prefer.

lopopolo (commented):
thanks for the feedback, new patch incoming.

lopopolo:
thanks for the feedback @Thomasdezeeuw. I have tried my best to address it and believe this PR is ready for another review.

Thomasdezeeuw (changes requested):
I'm sorry, but this is not at all what I wanted and I don't want to review this AI generated code any longer.

lopopolo:
> I'm sorry, but this is not at all what I wanted and I don't want to review this AI generated code any longer.

Hi Thomas, this is disappointing to hear. Thanks for your time.

lopopolo:
> I'm sorry, but this is not at all what I wanted and I don't want to review this AI generated code any longer.

I am trying to debug what was not acceptable here and I think it is because I did not construct the tests like this:

https://github.com/rust-lang/socket2/blob/642df44ae1eded922247300a83f8d60e159d9c9d/tests/socket.rs#L1602-L1607

I did not find this macro because https://github.com/rust-lang/socket2/pull/647#discussion_r3040744530 was not very specific in terms of which tests to match. Additionally, @Thomasdezeeuw had been providing dismissive feedback in response to AI assistance, so I wrote the commits 92be67d and 7ddafc0 by hand.

As someone familiar to Rust but new to this crate, _test macro_ most obviously meant `#[test]` and not a `macro_rules!` macro named `test`. I did find this in`CONTRIBUTING.md`:

> Tests should be added to `tests/socket.rs`, following (roughly) the same order in which the methods are defined on a type. At the bottom of this file it has a macro to create a simple get/set socket option test, more complex API however needs a manually written test.

but the docs do not match the reality of the test file; the macro is defined ~600 lines before the end of the file. When scanning the `tests/socket.rs` source in `vim`, I missed it. I attempted to find something approximating an `assert!` macro and found several assertion helpers like this one:

https://github.com/rust-lang/socket2/blob/642df44ae1eded922247300a83f8d60e159d9c9d/tests/socket.rs#L381-L390

which is what I used to model how I addressed the feedback. I also believed divergence from `std` was part of the bug here which warranted a dedicated test.

Based on my understanding now, I am guessing that the acceptable test would have been the following:

```rust
#[cfg(any(
    target_os = "dragonfly",
    target_os = "freebsd",
    target_os = "openbsd",
    target_os = "netbsd",
    target_os = "solaris",
    target_os = "illumos",
    target_os = "nto",
))]
test!(IPv4 multicast_ttl_v4, set_multicast_ttl_v4(258), 2);
```

I would have hoped that my good faith contributions could have been either:

- given constructive review feedback with the goal of getting the code mergable.
- merged and then fixed up by the maintainers with a followup commit.
- written directly by the maintainers in the first place after I reported the bug initially.
- engaged with in the context of the correctness issue this work identifies, reproduces, and fixes.

For the avoidance of doubt, this comment (and all of them on #647 and #646) are human-authored.

PR #650, opened 2026-04-12
Implement Send for MsgHdr and MsgHdrMut

sjthomason:
Allow MsgHdr and MsgHdrMut structs to be moved between threads. This is useful when using sendmsg() or recvmsg() calls on a non blocking socket from a spawned async task where the future would need to be Send.


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @sjthomason

PR #653, opened 2026-04-28
fix: fixes invalid libc symbols on horizon-os/n3ds

kore-signet:
HorizonOS (as used on the Nintendo 3DS) doesn't support all the calls/features currently expected. This PR just configures them all out, same as is done on wasi/espidf/redox.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -279,13 +279,13 @@ impl Type {
     pub const DCCP: Type = Type(sys::SOCK_DCCP);
 
     /// Type corresponding to `SOCK_SEQPACKET`.
-    #[cfg(all(feature = "all", not(any(target_os = "espidf", target_os = "wasi"))))]
+    #[cfg(all(feature = "all", not(any(target_os = "espidf", target_os = "wasi", target_os = "horizon"))))]
     pub const SEQPACKET: Type = Type(sys::SOCK_SEQPACKET);
 
     /// Type corresponding to `SOCK_RAW`.
     #[cfg(all(
         feature = "all",
-        not(any(target_os = "redox", target_os = "espidf", target_os = "wasi"))
+        not(any(target_os = "redox", target_os = "espidf", target_os = "wasi", target_os = "horizon"))
     ))]
     pub const RAW: Type = Type(sys::SOCK_RAW);
 }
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -384,7 +384,7 @@ impl RecvFlags {
     ///
     /// On Unix this corresponds to the `MSG_TRUNC` flag.
     /// On Windows this corresponds to the `WSAEMSGSIZE` error code.
-    #[cfg(not(target_os = "espidf"))]
+    #[cfg(not(any(target_os = "espidf")))]
```
Thomasdezeeuw: Is horizon missing here? If not, can you revert this?

src/socket.rs
```diff
@@ -21,10 +21,10 @@ use std::os::windows::io::{FromRawSocket, IntoRawSocket};
 use std::time::Duration;
 
 use crate::sys::{self, c_int, getsockopt, setsockopt, Bool};
-#[cfg(all(unix, not(target_os = "redox")))]
+#[cfg(all(unix, not(any(target_os = "horizon", target_os = "redox"))))]
```
Thomasdezeeuw: Nit: can you keep consistent with the others, i.e. horizon last?

src/socket.rs
```diff
@@ -632,7 +632,7 @@ impl Socket {
     /// <https://github.com/microsoft/Windows-classic-samples/blob/7cbd99ac1d2b4a0beffbaba29ea63d024ceff700/Samples/Win7Samples/netds/winsock/recvmsg/rmmc.cpp>
     /// for an example (in C++).
     #[doc = man_links!(recvmsg(2))]
-    #[cfg(all(unix, not(target_os = "redox")))]
+    #[cfg(not(any(target_os = "redox", target_os = "wasi", target_os = "horizon")))]
```
Thomasdezeeuw: This is incorrect, please use the following:
```suggestion
    #[cfg(all(unix, not(any(target_os = "redox", target_os = "horizon"))))]
```

Thomasdezeeuw (commented):
Mostly LGTM, couple of small changes requested.

Could you also add a check to the CI here: https://github.com/rust-lang/socket2/blob/602e78481b67d860d8747c7acfa40deb366c78d2/.github/workflows/main.yml#L67 ?

src/lib.rs
kore-signet: Reverted! The broader impl is cfg'd out regardless on Horizon, so it was redundant too.

src/socket.rs
kore-signet: Done :)

src/socket.rs
kore-signet: Oop, good catch!

kore-signet:
Latest commit fixes all the polish work! Seems like it made the docs build fail, though, so also fixing that real quick :)

kore-signet:
And docs build fixed! Thanks for the review :)

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @kore-signet

PR #655, opened 2026-05-25
Special case `assume_init` for `Bool` on Windows

### LINKED ISSUE #564: only_v6 panic on windows

ChrisDenton:
On Windows many socket options are documented as having the value "`DWORD` (boolean)". A `DWORD` is 4 bytes but when writing a boolean, often only one byte is actually written. While this is mostly consistent, it has been observed that using `getsockopt` with the `IPV6_V6ONLY` option can sometimes write 1 byte but other time write 4 bytes. So this PR allows us to accept either 1 or 4 bytes when the type of the option is `Bool`.

Fixes #564

### DIFF
--- a/src/sys/windows.rs
+++ b/src/sys/windows.rs
@@ -94,13 +94,53 @@ pub(crate) use windows_sys::Win32::Networking::WinSock::{
 pub(crate) const IPPROTO_IP: c_int = windows_sys::Win32::Networking::WinSock::IPPROTO_IP as c_int;
 pub(crate) const SOL_SOCKET: c_int = windows_sys::Win32::Networking::WinSock::SOL_SOCKET as c_int;
 
-/// Type used in set/getsockopt to retrieve the `TCP_NODELAY` option.
+// This is so we can special case MaybeUninit::assume_init for Bool.
+// See Bool for why.
+pub(crate) trait GetsockoptOutput: Sized {
+    unsafe fn assume_init(uninit: MaybeUninit<Self>, size: c_int) -> Self {
+        debug_assert_eq!(size as usize, size_of::<Self>());
+        uninit.assume_init()
+    }
+}
+
+impl GetsockoptOutput for i32 {}
+impl GetsockoptOutput for u32 {}
+impl GetsockoptOutput for IN_ADDR {}
+impl GetsockoptOutput for linger {}
+impl GetsockoptOutput for WSAPROTOCOL_INFOW {}
+
+/// Type used in getsockopt to retrieve options such as `TCP_NODELAY` or `IPV6_V6ONLY`.
 ///
 /// NOTE: <https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-getsockopt>
-/// documents that options such as `TCP_NODELAY` and `SO_KEEPALIVE` expect a
-/// `BOOL` (alias for `c_int`, 4 bytes), however in practice this turns out to
-/// be false (or misleading) as a `bool` (1 byte) is returned by `getsockopt`.
-pub(crate) type Bool = bool;
+/// documents that options such as `TCP_NODELAY` and `SO_KEEPALIVE` expect a "DWORD (boolean)".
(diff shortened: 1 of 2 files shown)


src/sys/windows.rs
```diff
@@ -94,13 +94,54 @@ pub(crate) use windows_sys::Win32::Networking::WinSock::{
...
+// This is so we can special case MaybeUninit::assume_init for Bool.
+// See Bool for why.
+pub(crate) trait AssumeInit: Sized {
+    unsafe fn assume_init(uninit: MaybeUninit<Self>, size: usize) -> Self {
+        debug_assert_eq!(size, size_of::<Self>());
+        uninit.assume_init()
+    }
+}
+
+impl<T: Copy> AssumeInit for T {}
```
ChrisDenton: Note that this fakes specialisation by not implementing `Copy` for the type (i.e. `Bool`) we want to special case. This is fine for our usage since we only actually need to move values and `Bool` is private.

Darksonn: What other types is this required by? Could we just explicitly implement it for the 3 or 4 other types?

ChrisDenton: It would currently need to be implemented for `i32`, `u32`, `IN_ADDR`, `linger` and `WSAPROTOCOL_INFOW`. Which seems fine. My only reluctance was that more things could be added in the future.

Darksonn: I would consider doing that and renaming this to `GetsockoptOutput` or similar.

ChrisDenton: Done!

src/sys/windows.rs
```diff
@@ -94,13 +94,58 @@ pub(crate) use windows_sys::Win32::Networking::WinSock::{
...
+        self.value == *other
+    }
+}
+
+impl GetsockoptOutput for Bool {
+    /// A wrapper around `MaybeUninit::assume_init` that will assume either 1 or 4 bytes are initialized.
+    ///
+    /// # Safety
+    ///
+    /// `size` must be either 1 or 4 and that many bytes must have been initialized.
```
Thomasdezeeuw: Nit: we don't need this documentation, the trait itself already explains it and the implementation does the rest.

src/sys/windows.rs
```diff
@@ -846,9 +895,8 @@ pub(crate) unsafe fn getsockopt<T>(socket: RawSocket, level: c_int, optname: i32
         SOCKET_ERROR
     )
     .map(|_| {
-        debug_assert_eq!(optlen as usize, mem::size_of::<T>());
         // Safety: `getsockopt` initialised `optval` for us.
-        optval.assume_init()
+        T::assume_init(optval, optlen as usize)
```
Thomasdezeeuw: Nit: do we want change the trait to accept an `c_int`? That way we don't need the cast here.

Thomasdezeeuw (approved):
This seems like a reasonable solution. Though I'm not super happy that we need to add this much code.

src/sys/windows.rs
ChrisDenton: Sure, that does make sense to me.

Thomasdezeeuw (approved):

Thomasdezeeuw:
GitHub Actions was stuck (what else is new), so I force pushed the commit to get it to start.

Thomasdezeeuw:
Thanks @ChrisDenton

PR #661, opened 2026-06-09
Define keepalive options for Emscripten

guybedford:
Extends the Emscripten sockets definitions to also include the keepalive options.


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @guybedford

PR #664, opened 2026-07-08
nuttx: type IovLen to c_ulong

wllenyj:
The `msg_iovlen` is the `unsigned long` in the Nuttx.

The origin define is as follows, in nuttx/include/sys/socket.h
```
struct msghdr
{
  FAR void *msg_name;
  socklen_t msg_namelen;
  FAR struct iovec *msg_iov;
  unsigned long msg_iovlen;
  FAR void *msg_control;
  unsigned long msg_controllen;
  unsigned int msg_flags;
};
```

### DIFF
--- a/src/sys/unix.rs
+++ b/src/sys/unix.rs
@@ -432,11 +432,12 @@ type IovLen = usize;
     target_os = "tvos",
     target_os = "watchos",
     target_os = "espidf",
-    target_os = "nuttx",
     target_os = "vita",
     target_os = "cygwin",
 ))]
 type IovLen = c_int;
+#[cfg(target_os = "nuttx")]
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @wllenyj.

You can also add NuttX to the check CI job here: https://github.com/rust-lang/socket2/blob/8f1c684890d09f5a8221d03e77dc63bbb9033c78/.github/workflows/main.yml#L61-L67, this way we can at least ensure it keep building in the future.

wllenyj:
@Thomasdezeeuw Thanks for your suggestion, it's usefull, but currently libc is not merged, so it can't compile for now.

